PR #58, opened 2014-08-26
Filters are now validated

smebberson:
This PR adds validation to the filters form controls. All fields are now required and new filters can't be included in a search, unless the user provides the necessary content.

@sandytrinh, this is ready for review. Thank you.


allanchau:
@smebberson Is this already in Linz? 🤔

PR #69, opened 2015-01-16
Concurrent editing

sandytrinh:
This PR gives the ability to review and resolve data conflict during concurrent editing. 

For example if 2 users are editing the same record at the same time, the first person to save the changes will be persisted in the database. When the second user attempting to save a change that was also edited by the first user, a modal will be opened prompting second user with instructions to resolve the conflict before he/she can save any changes.

### Todos
- [x] Add the ability to check if a record has been changed since user initiates an edit. This check should be performed when updating a record
- [x]  Add a modal screen with instruction for how to resolve a conflict. This modal screen should appear when user attempting to save a change and there is a record conflict
- [x] Add conflict message and options to resolve conflict 
 - [x] Text input
 - [x] Radio and checkboxes
 - [x] Select field
 - [x] Multiselect field
 - [x] Custom widget: google map picker
 - [x] Document array
 - [x] Textarea
- [x] Package this functionality into a Linz plugin
- [x] Create test suite for this new plugin


sandytrinh:
Based on recommendation from @smebberson we need to refactor the client side javascript that handles the conflict rendering into smaller modules and attach each individual javascript handler to the widget for each form field.

### Todos
- [ ] Integrate conflict handler into linz formtools widgets
- [ ] When handling the conflict on the client side, determine if ```{fieldName}HandleConflict()``` function exits first, if not execute the default ```{type}HandleConflict()```
- [ ] Add field attribute ```data-conflict-type``` to each form field

sandytrinh:
@smebberson This is ready for review.

### Testing

To test for data conflict between 2 users, open the same record in 2 different browser tabs. Edit and save the changes on the first tab. On the other tab, click 'Save' button, this should flag the changes you have made in the first tab.

Below are some suggestions for the test cases:
- Test that conflict changes are flagged for:
  - Text input (specifically, text, date and number fields)
  - Select field
  - MultiSelect field
  - Checkboxes
  - Radios
  - Document array
  - Textarea
- Test that data conflicts are flagged in a continuous loop upon each 'Save' if there is a change in the current version loaded in the browser and the version in the database. For example, if user A has merged all the data conflict upon the first conflict and did not immediately save the changes, any subsequent update for that particular record will flag as data conflict for user A the next time the user click on the 'Save' button, which means user A will need to perform any data conflicts highlighted again before the user can save any changes.

public/js/model/check-record-changes.js
```diff
@@ -0,0 +1,800 @@
...
+            }
+
+        }
+
+        yourChangeLabel = formatTextFieldLabel(yourChangeLabel);
+        theirChangeLabel = formatTextFieldLabel(theirChangeLabel);
+
+        if (!formField.parent().hasClass('has-change')) {
+
+            formField.wrap('<div class="input-group has-conflict has-change"></div>');
```
smebberson: @sandytrinh, do you think it's worth using mustache or something similar for this? Put the HTML into `<script type="template">` tags and rendering them first?

sandytrinh: Hey Scott, where would be a good spot to render the the ```script``` tag? As we wouldn't use this JS file to render the script tag. Perhaps in recordEdit.jade?

smebberson: Why wouldn't you have the code in here to render the script tag?

Let's say you have a template in HTML on the page:

```
<script id="template" type="x-tmpl-mustache">
Hello {{ name }}!
</script>
```

The following code will pass an object and render that template to HTML with you can then append to the DOM using `formField.after`...

```
var rendered = Mustache.render($('#template').html(), {name: "Luke"});
```

That doesn't seem like to much code? What do you think...

sandytrinh: Are you referring rendering ```script``` tag and it's content from the check-record-changes.js file e.g like building the  script tag with it's HTML content and insert it into the DOM on document.ready?

sandytrinh:
Added the following:
- [x] UI adjustment
 - [x] Make the conflict button background to orange so that it stands out
 - [x] Set min-wdith for the compare contents for document array and both table is even out
- [x] Bump up handlerbars (client side) to version 2.0.0 (there some new built in helpers that are required)
- [x] Move most of the HTML markup to a template and utilise handlebars to customise the content on client side

PR #71, opened 2015-02-09
Improving the record overview actions

smebberson:
At present, the record overview actions aren't as complete as they should be, they're lacking in a few areas. The following address these issues:

- [x] insert multiple actions into a drop-down menu (provides better support for devices)
- [x] allow an action to be disabled based on the record itself at runtime (the same as record actions work on the model index)
- [x] provide the ability to insert a [divider](http://getbootstrap.com/components/#dropdowns-divider) into the menu (also apply to model index record actions)
- [x] provide the ability to insert a [menu header](http://getbootstrap.com/components/#dropdowns-headers) (also apply to model index record actions)

## DSL

The following DSL can be applied to both model index record actions, and model overview actions.

### Standard menu item

```
{
	label: 'View raw data',
	action: 'dump'
}
```

### Standard menu item with disabling capability

```
{
	label: 'View raw data',
	action: 'dump',
	disabled: function (record, callback) {
		if (record.allowInformationDump) {
			return callback(null, false);
		}
		return callback(null, true, 'You are not allowed to preview all information for this record.');
	}
}
```

### Menu divider

```
{
	type: 'divider'
}
```

### Menu header

```
{
	label: 'View raw data',
	type: 'header'
}
```


sandytrinh:
@smebberson All works really well together. Only thing I'm not too sure is the placement of the action button on the overview screen. When you have a group buttons, it shows a dropdown 'Actions' which makes sense for it to appear on the left hand side (next to Edit) but when you have only one action button e.g. 'Send welcome email' it doesn't look as great and for some models the actions will be different. It looks a bit inconsistent. What are your thoughts on this?

smebberson:
@sandytrinh, good point. I've moved the actions to go into between the Delete button (if it exists) and the View All button (if it exists). What do you think, ready to merge?

sandytrinh:
Great! Please merge!

PR #72, opened 2015-02-11
Model edit improvements

sandytrinh:
This PR fixes the following:

- [x] Formtools form generator
 - [x] Update ```generateFormFromModel()``` function to exclude fields where ```field.visible === false``` first and only process visible fields
 - [x] Enable the form builder to pass the error back to the router which can be pushed to express error handler. Currently any errors occurred in the form builder is concealed.
 - [x] Throw error if a form field is of type ```documentArray``` and it's not bind to the Linz embeddedDocument plugin
- [x] Updated express error handler to output the error stack to console

### DIFF
--- a/lib/formtools/form.js
+++ b/lib/formtools/form.js
@@ -62,6 +62,10 @@ var generateFormFromModel = exports.generateFormFromModel = function (m, schemaF
             // handle list types
             function (callback) {
 
+                if (field.visible === false) {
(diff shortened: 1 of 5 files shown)


smebberson:
@sandytrinh, this is a nice addition. Tidies up some of the internals within Linz. Great one. Please merge.

PR #73, opened 2015-02-11
Added a route to output a JSON string for configs and model records.

smebberson:
As per the title, this allows you to easily access the raw JSON data for a model record, or a config. Simply replace `overview` at the end of a record overview with `json` to see the output.

I've decided not to create a model overview action 'button' for this, until we have the capability to lock down actions to certain user groups. I don't think everyone should have a button on the overview screen for this, although I don't mind them having access to the raw data.


sandytrinh:
@smebberson Looks good, merge away!

PR #74, opened 2015-02-11
Pass mongoose objects to model index renderers, rather than object literals

smebberson:
One of the features of mongoose is the ability to have model statics and methods. It's great and allows you to contextify additional functionality to a model quite easily.

When rendering the model index page, we turn the mongoose object into an object literal so that we can attach extra properties to the object (mongoose objects don't allow this). The extra properties are used when rendering the model index page in jade.

This change provides mongoose objects to model index renderers, and provides object literals with extra properties to jade for rendering.

@sandytrinh, this PR is ready for review.

### DIFF
--- a/middleware/_modelIndex.js
+++ b/middleware/_modelIndex.js
@@ -279,6 +279,9 @@ module.exports = function  (req, res, next) {
                     // loop through each record
                     async.each(Object.keys(records), function (index, recordDone) {
 
+                        // store the rendered content into a separate property
+                        records[index]['rendered'] = {};
+
                         // loop through each column
                         async.each(Object.keys(req.linz.model.grid.columns), function (column, columnDone) {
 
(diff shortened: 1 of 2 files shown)


sandytrinh:
@smebberson All is working well!

This is a bit of nitpicking and I know some of the code were there before you made the change, but do you think the reference to ```index``` can be changed to ```fieldName```. One of the many example where this variable appears is ```mongooseRecords[index]```.

```index``` tends to represent a number (e.g. element in an array) whereas in this case, it is referencing a key name so it makes more sense to have the variable as ```fieldName``` which is also consistent with the code above in one of the asyn function.

What do you think?

smebberson:
@sandytrinh, this is a great question and I don't think it's nit picking. Writing code that is easy to understand is very important!

As for this instance, `mongooseRecords` is indeed an `Array`. Using `Object.keys(mongooseRecords)` will provide you an `Array` of numbers. So we're looping over numbers and therefore `index` is appropriate I think.

The reason we're looping over array numbers is so that it's easy to swap the reference between records and `mongooseRecords`. The `records` `Array` holds the object literal version of the records, and the `Array` `mongooseRecords` holds the mongoose object version of the records.

sandytrinh:
Oh I didn't realise that ```mongooseRecords``` is an array. I just assumed that it is an object because of ```Object.keys(mongooseRecords)```. Fair enough!

Please merge away!

PR #75, opened 2015-02-19
Fixes model index pages without filters

smebberson:
The data grid sorting/pagination on model index pages without filters was essentially broken. These changes make them work again.

@sandytrinh, this is just a small one and read for review.

### DIFF
--- a/lib/helpers-models.js
+++ b/lib/helpers-models.js
@@ -22,9 +22,9 @@ exports.loadModels = function loadModels (dir, cb) {
     try {
         files = fs.readdirSync(dir);
     } catch (e) {
-        debugConfigs('Error: Unable to load models from directory, %s', dir);
+        debugModels('Error: Unable to load models from directory, %s', dir);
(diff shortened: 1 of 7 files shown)


sandytrinh:
Hi @smebberson 

It looks like when a filter is not present for a model, if you are currently on page 2 and changed the total records displayed per page to e.g. 50, the page refreshed and shows blank!

I was able to replicate this use on mmsUser.

smebberson:
@sandytrinh,

Thank for your feedback. I've resolved this issue by reseting the page index whenever the page size changes. I've also made an update so that the page size controls are always rendered.

Can you please take another look?

sandytrinh:
Tested this on my user model (without filter settings in the grid DSL) and it still not reseting the page index to one when changing the page size. Performed this test on another model and the result is the same.

allanchau:
Pretty sure this has been fixed somewhere along the line, we should probably open a new PR if it is still an issue anyway.

PR #77, opened 2015-03-10
Multiple exports for a model

smebberson:
Linz's default model export functionality is pretty good, but there are times when you just want to do something different.

This PR provides the ability to either completely overwrite the default export capability, or supplement it with another export option for your users.

Todo
-------

- [x] Update Linz to accept either an object or an array
- [x] Update the model index to show a drop-down list if there are multiple exports
- [x] Ensure that custom export routines can't use `action` with a value of `export` (this has been reserved for Linz)
- [ ] Update the tests to ensure we have test coverage for this

Usage
---------

To use this functionality, add an `export` property to your model's `grid` property:

```
    grid: {
        export: [
            {
                label: 'Custom export',
                action: 'url/endpoint'
            }
        ]
    }
```

To use a custom export, along with Linz's default export:

```
    grid: {
        export: [
            {
                label: 'Export',
                exclusions: 'fields,to,exclude'
            },
            {
                label: 'Custom export',
                action: 'url/endpoint'
            }
        ]
    }
```

Providing an export definition with an `action` value of `export` will error, for example:

```
    grid: {
        export: [
            {
                label: 'Export',
                action: 'export'
            }
        ]
    }
```

### DIFF
--- a/lib/formtools/plugins/plugins-helpers.js
+++ b/lib/formtools/plugins/plugins-helpers.js
@@ -91,17 +91,7 @@ var defaults = function (options, isEmbeddedDocument) {
     options.grid.recordActions = options.grid.recordActions || [];
 
     // defaults to []
-    options.grid.export = options.grid.export || [];
-
-    options.grid.export.label = options.grid.export.label || 'Export';
-
-    options.grid.export.action = options.grid.export.action || 'export';
-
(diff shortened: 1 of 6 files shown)


smebberson:
This PR has been superseded by #78.

PR #78, opened 2015-03-11
Custom exports

smebberson:
This PR is stage 1 of a two stage improvement to Linz in which custom exports will be supported. Stages are broken down into:

1. the ability to define a custom export that doesn't require a modal UI step (as per the default Linz export)
1. the ability to define a custom export that does require a modal UI step

Stage 2 will be implemented in a separate PR.

Linz's default model export functionality is pretty good, but there are times when you just want to do something different. This PR provides the ability to either completely overwrite the default export capability, or supplement it with another export option for your users.

Todo
-------

- [x] Update Linz to accept either an object or an array for the grid.export property
- [x] Update the model index to show a drop-down list if there are multiple exports
- [x] Ensure that custom export routines can't use `action` with a value of `export` (this has been reserved for Linz)
- [ ] Update the tests to ensure we have test coverage for this

Usage
---------

To use this functionality, add an `export` property to your model's `grid` property:

```
    grid: {
        export: [
            {
                label: 'Custom export',
                action: 'url/endpoint'
            }
        ]
    }
```

To use a custom export, along with Linz's default export:

```
    grid: {
        export: [
            {
                label: 'Export',
                exclusions: 'fields,to,exclude'
            },
            {
                label: 'Custom export',
                action: 'url/endpoint'
            }
        ]
    }
```

Providing an export definition with an `action` value of `export` will error, for example:

```
    grid: {
        export: [
            {
                label: 'Export',
                action: 'export'
            }
        ]
    }
```

### DIFF
--- a/lib/formtools/plugins/plugins-helpers.js
+++ b/lib/formtools/plugins/plugins-helpers.js
@@ -91,17 +91,7 @@ var defaults = function (options, isEmbeddedDocument) {
     options.grid.recordActions = options.grid.recordActions || [];
 
     // defaults to []
-    options.grid.export = options.grid.export || [];
-
-    options.grid.export.label = options.grid.export.label || 'Export';
-
-    options.grid.export.action = options.grid.export.action || 'export';
-
-    // defaults to false
-    options.grid.export.enable = (options.grid.export.enable === true);
-
-    options.grid.export.exclusions = options.grid.export.exclusions || '_id,dateCreated,dateModified,createdBy,modifiedBy';
-
+    options.grid.export = exportDefault(options.grid.export) || [];
 
     /**
     * overview defaults
@@ -249,6 +239,50 @@ var filtersDefault = function (filters) {
     });
 
     return filters;
+
+};
+
+var exportDefault = function (exports) {
+
+    if (!exports) {
+        return undefined;
+    }
+
+    var validatedExports = [];
+
+    // linz will accept an object, or an array of objects
+    if (!Array.isArray(exports)) {
+        exports = [exports];
+    }
+
+    // loop through each export, and default some values
+    exports.forEach(function (exp) {
+
+        // we must have a label, otherwise there is not enough information
+        if (!exp.label) {
+            return;
+        }
+
+        if (exp.action && exp.action === 'export') {
+            throw new Error('You can not define a model export with an action property of export, this is reserved for Linz.');
+        }
+
+        var expObj = {
+                label: exp.label,
+                action: exp.action || 'export',
+                enabled: !(exp.enable === true),
+                exclusions: exp.exclusions || '_id,dateCreated,dateModified,createdBy,modifiedBy'
+            };
+
+        // Linz's built in action is called export
+        expObj.custom = expObj.action !== 'export';
+
+        // push this into the list of exports
+        validatedExports.push(expObj);
+
+    });
+
+    return validatedExports;
 
 };
 
--- a/middleware/modelExport.js
+++ b/middleware/modelExport.js
@@ -109,12 +109,12 @@ var modelExportHelpers = function modelExportHelpers (req, res) {
 
         getFilters: function getFilters (cb) {
 
-            if (!req.body.modelQuery.length) {
+            if (!req.body.filters.length) {
                 return cb(null, {});
             }
 
             var Model = linz.get('models')[req.body.modelName],
-                form = JSON.parse(req.body.modelQuery);
+                form = JSON.parse(req.body.filters);
 
             // check if there are any filters in the form post
             if (!form.selectedFilters) {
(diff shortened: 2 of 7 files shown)


smebberson:
@sandytrinh, I'd like to merge this. I know all of the todo's haven't been completed, but it's a little hard to to unit testing with Linz at the moment (I'll get started on a Vagrant setup for Linz straight away).

Other than that though, it's all working. Could you please review...

lib/formtools/plugins/plugins-helpers.js
```diff
@@ -252,6 +242,50 @@ var filtersDefault = function (filters) {
 
 };
 
+var exportDefault = function (exports) {
+
+    if (exports === undefined || exports === null) {
+        return undefined;
```
sandytrinh: Is there a difference between L247 and ```if (!exports)```?

smebberson: @sandytrinh, good point. I've fixed (through a forced update). You'll need to pull down the repo to see the update.

middleware/modelExport.js
```diff
@@ -298,7 +328,18 @@ module.exports = {
         asyncFn.push(helpers.getForm);
         asyncFn.push(helpers.getGrid);
 
-        async.waterfall(asyncFn, function (err, filters, form, grid) {
+        // get the actual export object
```
sandytrinh: Should the code on L318 be superseded due to these new changes for export?

lib/formtools/plugins/plugins-helpers.js
sandytrinh: Did you amend your last commit with this change?

smebberson: Yes, I amended my commit and force pushed. I was naughty. I just wanted to give it a try. Whoops.

middleware/modelExport.js
smebberson: I thought that was still a good option? Allowing them to provide an export function to replace absolutely everything...

Or do you think this multiple exports thing is good enough we shouldn't allow a complete override... I'm thinking maybe you're right...

lib/formtools/plugins/plugins-helpers.js
```diff
@@ -252,6 +242,50 @@ var filtersDefault = function (filters) {
...
+        }
+
+        if (exp.action && exp.action === 'export') {
+            throw new Error('You can not define a model export with an action property of export, this is reserved for Linz.');
+        }
+
+        var expObj = {
+                label: exp.label,
+                action: exp.action || 'export',
+                enabled: !(exp.enable === true),
```
sandytrinh: What purpose will ```enabled``` serve here that now export is not defined by default? (L248)

smebberson: Enabled simply allows you to disable an export. I thought it was good a option rather than just having to delete the code... you don't like?

middleware/modelExport.js
sandytrinh: I think the multi exports (with stage 2 implemented) will give them the ability to write their own if the default export don't meet their needs.

I'm assuming you can define an export button without having to use the default Linz export. I'm basing my assumption on your example DSL below:

```
    grid: {
        export: [
            {
                label: 'Custom export',
                action: 'url/endpoint'
            }
        ]
    }
```

lib/formtools/plugins/plugins-helpers.js
sandytrinh: I see, no wonder I had to merge mine with the latest from your remote branch. I deleted my local version and pull yours down again as a fresh branch to avoid the unnecessary merge :smile:

middleware/modelExport.js
smebberson: Yeah, okay, I agree. I'll get rid of it.

To use the default Linz export, the least amount code to achieve it is:

```
grid: {
    export: {
        label: 'Export'
    }
}
```

And then override the default export, the least amount of code to achieve it is:

```
grid: {
    export: {
        label: 'Custom export',
        action: 'url/endpoint'
    }
}
```

You can then use any combination of the above, by turn the export object, into an Array.

lib/formtools/plugins/plugins-helpers.js
sandytrinh: Just thinking of a scenario where you would use that? Since, you would define an export when you need it and when would you define an export and disabled it?

smebberson: Not super useful now, no. But what about when we have permissions. Being able to easily enable/disable an export would be handy I think...

sandytrinh: Good point with the permission example. Ok let's keep it!

middleware/modelExport.js
sandytrinh: When I used the example below, clicking on the 'Custom export' button opens up a modal. This is the same when the export is an array.

```
grid: {
    export: {
        label: 'Custom export',
        action: 'url/endpoint'
    }
}
```

lib/formtools/plugins/plugins-helpers.js
smebberson: Sorry. I didn't realise it would have that affect. I guess that is why you don't do it! This will help in the future though:

```
git fetch
git reset --hard origin/{branch-name-or-master}
```

middleware/modelExport.js
smebberson: Ooooh. That's a bug. Thanks. I'll fix.

sandytrinh:
@smebberson Beside the minor bug with the custom export opening in a modal, everything else is working very well! Good work!

sandytrinh:
Also will linz provide the ability to overwrite the default linz export 'transform' function?

middleware/modelExport.js
smebberson: That should be fixed now :)

smebberson: Model.export functionality has been removed :)

sandytrinh:
@smebberson This is all good! Please merge!

smebberson:
I'll leave the export transform override to another PR :)

PR #79, opened 2015-03-12
Grid toolbar

sandytrinh:
This PR adds the ability the following features:
- [x] Ability to add toolbar items to the model index page
- [x] Ability to use api functions for grid formtools

### Grid toolbar items

The following DSL will allow you to provide your own HTML markup to create a toolbar that appear on the top of the page next to your 'Create {modelName}' button.

```
{
    grid: {

        toolbarItems: [
            {
                renderer: function (req, res, cb) {                
                    // code to render your toolbar items
                    var html = '';

                    return cb(null, html);
                }
            }
        ]

    }
}
```

Linz will not provide any basic renderer as the toolbar is a custpm UI that user can create independently.

### Formtolls grid api

This PR is the first step to api-rise linz. The first obvious step is to move functions for the model index view into it's own api. This api is accessible via ```linz.api.formtools.grid```.

Currently ```linz.api.formtools.grid``` has 5 public functions, there are:
- getSettings
- getToolbarItems
- getFilters
- getSearchFilters
- getActiveFilters

You can see a working example of these function in ```middleware/_modelIndex.js``` file.

Due to the addition of linz.api.formtools, the existing linz.api was refactored to linz.api.admin. For this reason, this :warning:**this PR contains breaking changes.**

## Breaking change updates (30/3/2015).

The latest commit included another breaking changes. We have updated the ```linz.get('models').formtools``` to ```linz.get('models').linz.formtools``` to ensure model data are in the linz namespace.


lib/api/formtools/grid.js
```diff
@@ -0,0 +1,263 @@
+var linz = require('../../../');
+    async = require('async'),
+    clone = require('clone');
+
+(function  () {
```
smebberson: How come this is a self executing function?

lib/api/formtools/grid.js
```diff
@@ -0,0 +1,263 @@
...
+
+    function getSettings(modelName, cb) {
+
+        if (!modelName || typeof(modelName) !== 'string' || !modelName.length) {
```
smebberson: This could be rewritten to `if (typeof modelName !== 'string' || !modelName.length) {`

lib/api/formtools/grid.js
```diff
@@ -0,0 +1,263 @@
...
+            throw new Error('modelName is required and must be of type string and cannot be emptied.');
```
smebberson: Should read "modelName is required and must be of type String and cannot be empty."

lib/api/formtools/grid.js
```diff
@@ -0,0 +1,263 @@
...
+                return cb(err);
+            }
+
+            return cb(err, clone(settings));
+
+        });
+
+    }
+
+    function getToolbarItems (req, res, modelName, cb) {
```
smebberson: I think we should both install https://atom.io/packages/docblockr and make sure that we use this. I think this code would benefit from some basic explanations of the functions. What do you think?

lib/api/formtools/grid.js
```diff
@@ -0,0 +1,263 @@
...
+                return cb(err);
+            }
+
+            if (!settings.toolbarItems.length) {
+                return cb(null);
+            }
+
+            async.each(settings.toolbarItems, function (item, itemDone) {
+
+                if (!item.renderer) {
```
smebberson: Do you think we should throw an error here? A toolbarItem without a renderer is essentially useless. It might help people implementing a toolbarItem to know when they've done something wrong. It will always be an implementation issue rather than a runtime issue too.

lib/api/formtools/grid.js
```diff
@@ -0,0 +1,263 @@
...
+            });
+
+        });
+
+    }
+
+    function getSearchFilters(fieldNames, data, modelName, cb) {
+
+        if (!fieldNames || !Array.isArray(fieldNames) || !fieldNames.length) {
+            throw new Error('fieldNames is required and must be of type array and cannot be emptied.');
```
smebberson: Should read "fieldNames is required and must be of type Array and cannot be empty."

lib/api/formtools/grid.js
```diff
@@ -0,0 +1,263 @@
...
+
+                return cb(filterErr, settings.filters);
+
+            });
+
+        });
+
+    }
+
+    function getSearchFilters(fieldNames, data, modelName, cb) {
```
smebberson: What is a search filter? Is this function used to generate the filters that will be run against he database?

lib/api/formtools/grid.js
```diff
@@ -0,0 +1,263 @@
...
+            });
+
+        });
+
+    }
+
+    function getActiveFilters (selectedFilters, data, modelName, cb) {
+
+        if (!selectedFilters || !Array.isArray(selectedFilters) || !selectedFilters.length) {
+            throw new Error('selectedFilters is required and must be of type array and cannot be emptied.');
```
smebberson: This should read "selectedFilters is required and must be of type Array and cannot be empty."

lib/api/formtools/grid.js
```diff
@@ -0,0 +1,263 @@
...
+                return cb(filterErr, filters);
+            });
+
+        });
+
+    }
+
+    function getActiveFilters (selectedFilters, data, modelName, cb) {
+
+        if (!selectedFilters || !Array.isArray(selectedFilters) || !selectedFilters.length) {
```
smebberson: I don't think you need the falsy check at the start for this to work...

lib/api/formtools/grid.js
```diff
@@ -0,0 +1,263 @@
...
+            throw new Error('selectedFilters is required and must be of type array and cannot be emptied.');
+        }
+
+        if (!modelName || typeof(modelName) !== 'string' || !modelName.length) {
```
smebberson: This could be rewritten to `if (typeof modelName !== 'string' || !modelName.length) {`

lib/api/formtools/grid.js
```diff
@@ -0,0 +1,263 @@
...
+            throw new Error('modelName is required and must be of type string and cannot be emptied.');
+        }
+
+        if (!data || typeof(data) !== 'object') {
```
smebberson: I'm assuming that you want an object here, and not an Array. If so, this could be written as `if (!(data instanceof Object && !Array.isArray(data))) {`. Using instanceof is a good choice here and protects against null objects. The code as it current standards will accept Array's to be passed in.

smebberson:
This is just something to consider, but I was thinking about the re-namespacing of the `api.getAdminLink` function into `api.admin.getAdminLink`. I think it's a good re-namespacing.

However, do you think we should call it the `url` library, instead of `admin`? Admin just feels a little too generic, I guess because Linz in its entirety is used to build admins. This library is specific to urls.

I know it's kind of outside of the scope of this PR, but since we're moving it around anyway?

lib/api/formtools/grid.js
```diff
@@ -0,0 +1,263 @@
...
+            throw new Error('data is required and must be of type object.');
+        }
+
+        var model = linz.get('models')[modelName],
+            activeFilters = {};
+
+        async.waterfall([
```
smebberson: This pattern is used a lot:

```
async.waterfall([

            function (callback) {
                return getSettings(modelName, callback);
            }

        ], function (err, settings) {
```

Do you think it would be better to write the function once, and reference it here multiple times? I'm not sure... I'm just wondering if we can lay this out a little better...

lib/api/formtools/grid.js
```diff
@@ -0,0 +1,263 @@
...
+
+            if (err) {
+                return cb(err);
+            }
+
+            if (!Object.keys(settings.filters).length) {
+                return cb(null);
+            }
+
+            async.each(selectedFilters, function (fieldName, filtersDone) {
```
smebberson: This is being used to populate `activeFilters`. Perhaps it would be better to use `async.filter`, chained with `async.map` and then return the result to the callback?

lib/api/formtools/grid.js
```diff
@@ -0,0 +1,263 @@
...
+
+            }, function (itemErr) {
+                return cb(itemErr, settings.toolbarItems);
+            });
+
+        });
+
+    }
+
+    function getFilters (modelName, cb) {
```
smebberson: This gets filters and renders them, do you think it needs to be broken down? `getFilters` would simply get and `renderFilters` would get and render? I don't know... can you think of any instances in which on would want to get without rendering?

smebberson:
Overall this is looking and working great. Just a few questions regarding syntax and methods used to achieve things. I guess I'm just looking to set the scene with this code as to how we can best do things now and into the future.

lib/api/formtools/grid.js
sandytrinh: This is to allow internal access to the functions within the scope of this file using self executable function to encapsulate this scope and only expose the functions that are accessible outside of this scope.

lib/api/formtools/grid.js
sandytrinh: Done!

lib/api/formtools/grid.js
sandytrinh: Done!

lib/api/formtools/grid.js
sandytrinh: Yep agree. I've added the comments.

lib/api/formtools/grid.js
sandytrinh: I've updated this to return an error instead
```
return itemDone(new Error('Renderer function is missing for toolbar item: ' + require('util').inspect(item) + '. Check your \'' + modelName + '\' model for more details.'));
```

lib/api/formtools/grid.js
sandytrinh: I'll removed the self executive function since all variables defined within the module are local.

lib/api/formtools/grid.js
sandytrinh: I've renamed:
- ```getFilters``` to ```renderFilters```
- ```getToolbarItems``` to ```renderToolbarItems```
- ```getSearchFilters``` to ```renderSearchFilters```

so the function names are more inline with the function itself.

lib/api/formtools/grid.js
sandytrinh: Have renamed this to ```renderSearchFilters``` instead.

lib/api/formtools/grid.js
sandytrinh: Done!

lib/api/formtools/grid.js
```diff
@@ -0,0 +1,263 @@
+var linz = require('../../../');
+    async = require('async'),
+    clone = require('clone');
+
+(function  () {
+
+    function getSettings(modelName, cb) {
```
smebberson: As per our discussion, it would be good to get rid of this and start making use of the model bootstrapping we have in place which places the grid settings in `model.linz.formtools.grid`.

lib/api/formtools/grid.js
sandytrinh: I've applied the changes

lib/api/formtools/grid.js
sandytrinh: I've applied the changes

lib/api/formtools/grid.js
sandytrinh: I've applied the changes

lib/api/formtools/grid.js
sandytrinh: I've applied the changes

lib/api/formtools/grid.js
sandytrinh: I've refactored this to utilise the linz api for grid settings.

lib/api/formtools/grid.js
sandytrinh: This is not applicable since the end result is an object rather than an array.

sandytrinh:
Hey @smebberson,

In regards to the ```url`` library, I actually did thought about changing it to the ```url``` api instead of ```admin``` but I thought at the time perhaps it would include another more admin specific api functions. But yeah we definitely can change it. Do you want it in ```linz.api.url``` namespace or ```linz.api.admin.url```?

I should in linz's perspective, it would only deal with admin URLs?

smebberson:
Hey @sandytrinh,

Yeah, I think `linz.api.url` would be a good choice. Nice one.

lib/api/formtools/grid.js
sandytrinh: Done!

sandytrinh:
@smebberson All suggestions are implemented, shall I merge?

smebberson:
@sandytrinh, please merge!

PR #83, opened 2015-06-04
Custom Passport configuration

smebberson:
This PR provides the ability to override the default Linz passport configuration with your own. It allows you to customise the way in which Linz finds a user in the database, serialisers and deserialises a user.


smebberson:
@sandytrinh, use this branch to test a custom Passport configuration.

PR #84, opened 2015-06-22
Improves linz modal support

sandytrinh:
This PR improves linz modal support where you can load a bootstrap modal by setting ```data-linz-modal="true"``` attribute in your html ```a``` or ```button``` tag.

linz modal also includes form validation by adding ```data-linz-validation="true"``` to the ```form``` tag in the modal HTML.

@smebberson This is ready for review

### DIFF
--- a/public/js/ui.js
+++ b/public/js/ui.js
@@ -121,26 +121,36 @@ if (!linz) {
         });
 
         // add ability to open URL in a modal
-        $('[data-linz-modal]').click(function () {
+        $('[data-linz-modal]').click(function (event) {
 
-            if ($(this).attr('data-linz-disabled') === 'true') {
+            event.preventDefault();
+
+            var button = $(this);
+
+            if (button.attr('data-linz-disabled') === 'true') {
                 return;
             }
 
-            var queryObj = $(this),
-                url = queryObj[0].nodeName === 'BUTTON' ? queryObj.attr('data-href') : queryObj.attr('href');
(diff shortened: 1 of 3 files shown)


public/js/ui.js
```diff
@@ -120,6 +120,30 @@ if (!linz) {
...
+        $('[data-linz-modal]').click(function () {
+
+            if ($(this).attr('data-linz-disabled') === 'true') {
+                return;
+            }
+
+            var queryObj = $(this),
+                url = queryObj[0].nodeName === 'BUTTON' ? queryObj.attr('data-href') : queryObj.attr('href');
+
+            $('#linzModal').modal().load(url, function () {});
```
smebberson: Seeing as the `load` function is provided by jQuery, let's support the entire api which is `.load( url [, data ] [, complete ] )`. As per the `load` function, `data` and `complete` should be optional.

sandytrinh: @smebberson, we have agreed that including ```data``` and ```complete``` don't provide any benefit for this feature. However, I've added the ability to define a ```callback``` function that will execute after the modal is shown. Below are 2 examples of the modal DSL:

```
# modal with a callback
modal : {
    callback: 'jsFunctionName'
}

# modal with no callback
modal : true
```

sandytrinh:

smebberson:
@sandytrinh, looks good. Please merge!

PR #87, opened 2015-06-25
Local date time cell renderer

sandytrinh:
This PR adds the ability to render local datetime using client side scripting.
Linz will now allow customisation for:
- date format
- datetime format
- localise date by default

This configuration can change when initialising the application. See linz ```default.js``` for the default values.

This PR also includes 3 other date renderers, they are:
- ```dateTimeRenderer```
- ```localDateRenderer```
- ```localDateTimeRenderer```

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -30,6 +30,9 @@ var defaults = module.exports = {
     'page size': 20,
     'page sizes': [20,50,100,200],
 	'cookie secret': 'oST1Thr2s/iOAUgeK4yuuA==',
-	'requires login': true
+	'requires login': true,
+	'date format': 'ddd DD/MM/YYYY',
+	'datetime format': 'ddd DD/MM/YYYY h:mm A',
+	'localise time': false
 
 };
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -31,8 +31,8 @@ var defaults = module.exports = {
     'page sizes': [20,50,100,200],
 	'cookie secret': 'oST1Thr2s/iOAUgeK4yuuA==',
 	'requires login': true,
-	'date format': 'ddd DD/MM/YYYY',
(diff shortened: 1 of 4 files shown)


smebberson:
@sandytrinh, how is the `localDateRenderer` and `localDateTimeRenderer` used on the client-side? Because ISO 8601 isn't really a localised date format?

Have you seen http://momentjs.com/docs/#/i18n/instance-locale/? Would it be better to use something like that? I'm not really sure how it's being used...

lib/formtools/renderers-cell.js
```diff
@@ -5,7 +5,28 @@ var moment = require('moment'),
...
+};
+
+var dateTimeRenderer = module.exports.dateTime = function dateTimeRenderer (date, record, fieldName, model, callback) {
+    return callback(null, moment(date).format(linz.get('datetime format')));
+};
+
+var localDateRenderer = module.exports.localDate = function localDateRenderer (date, record, fieldName, model, callback) {
+
+    if (date instanceof Date) {
+        return callback(null, '<span data-linz-localise-date data-linz-date-format="' + linz.get('date format') + '">' + date.toISOString() + '</span>');
```
smebberson: Rather than span, would it be better to use the [`time`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) element?

lib/defaults.js
```diff
@@ -30,6 +30,9 @@ var defaults = module.exports = {
     'page size': 20,
     'page sizes': [20,50,100,200],
 	'cookie secret': 'oST1Thr2s/iOAUgeK4yuuA==',
-	'requires login': true
+	'requires login': true,
+	'date format': 'ddd DD/MM/YYYY',
```
smebberson: Could we please use locale based format strings, such as `llll` here?

PR #88, opened 2015-07-06
The PR provides the ability to change the login form

smebberson:
The two main components of this PR are:

- provide a customer login form renderer
- update Linz so that you can pass in an initialised Passport instance (just like you can an Express, or Mongoose instance)

Using these two updates, I've been able to completely rewrite the login that a Linz instance to use something completely different, such as a single sign-on setup.


linz.js
```diff
@@ -468,9 +474,15 @@ Linz.prototype.bootstrapExpress = function (cb) {
...
 	}
 
-	// initialize passport
-	this.app.use(this.get('admin path'), passport.initialize());
-	this.app.use(this.get('admin path'), passport.session());
+	// setup connect-flash
+	this.app.use(flash());
+
+	// configure passport
+	libPassport(this.passport);
```
sandytrinh: How does this works when there's a custom passport already configured with `serializeUser` and `deserializeUser` (like in the rascamp example)?

It looks like the default function `serializeUser` and `deserializeUser` in `lib/passport.js` are ignored?

sandytrinh:
This PR looks good! Once the issue for https://github.com/flashmedia/rascampaigner/pull/3 is resolved, this PR can be merged!

linz.js
smebberson: Yeah, that is a good point. I think libPassport needs to be updated so that it doesn't create a `serializeUser` or `deserializeUser` if they already exist.

PR #89, opened 2015-07-09
Implementation of a basic permission functionality

smebberson:
This PR is the implementation of the functionality as discussed in #8.

- [x] provide the ability to add to Linz's navigation structure (including a cache)
- [x] create a no-op function that will provide access to everything as a default
- [x] exclude permissions code from executing, if the no-op function has been overwritten
- [x] implement the ability to control predefined permissions for **model routes**:
    - [x] model create
    - [x] model edit
    - [x] model delete
    - [x] model json
    - [x] model export
    - [x] model overview
- [x] implement the ability to control **model route** permissions on custom model actions
- [x] implement the ability to control predefined permissions for **model views**:
    - [x] model create
    - [x] model edit
    - [x] model delete
    - [x] model json
    - [x] model export
    - [x] model overview
- [x] implement the ability to control **model view** permissions for custom model actions
- [x] implement the ability to control predefined permissions for **config routes**:
    - [x] config overview
    - [x] config reset
    - [x] config edit
    - [x] config json
- [x] implement the ability to control predefined permissions for **config views**:
    - [x] config overview
    - [x] config reset
    - [x] config edit
    - [x] config json
- [x] cache data per user session for:
 - [x] navigation
 - [x] model permissions
 - [x] configs permissions
- [x] provide a public middleware `linz.middleware.exclusions`


linz.js
```diff
@@ -663,18 +645,35 @@ Linz.prototype.buildNavigation = function (cb) {
 
         },
 
-        function (done) {
+		// this function will run and return either a no-op function,
+		// or a function that will execute the 'navigation configuration' function and override
+		// nav with the customisation as returned to the callback
+		(function () {
```
smebberson: This is a rethink of an issue we keep coming up against. The issue is:

__*how best do we structure a series of async functions, in which not all of them are required under all conditions*__

This implementation uses a self executing function to determine if a function is required, or a no-op should be used in place of performing any actions...

sandytrinh: Hi @smebberson,

Interesting concept! The self-executing function determines if an `op` is required for that particular function, what about when we want to skip all of the remaining functions? For example, if there are 5 more remaining functions in the series we want to skip, it would have to execution all of the remaining function to get to the end of the series?

sandytrinh:
Hi @smebberson,

I've completed the remaining tasks you had for this PR. Is there anything else we need to add for this particular feature?

If not, can you please review?

sandytrinh:

lib/api/configs.js
```diff
@@ -0,0 +1,80 @@
...
+
+        function (perms, callback) {
+
+            var permissionsObj = {};
+
+            permissions.forEach(function (permission) {
+                permissionsObj[permission] = perms.indexOf(permission) >= 0;
+            });
+
+            req.session.configs = req.session.configs || {};
```
smebberson: This line seems redundant?

lib/api/configs.js
```diff
@@ -0,0 +1,80 @@
...
+ * @param {Function} cb        Callback function
+ * @return {Object}
+ */
+function getPermissions(req, configName, cb) {
+
+    if (!linz.get('disabled config cache') && req.session.configs) {
+        return cb(null, req.session.configs);
+    }
+
+    var config = get(configName);
```
smebberson: There doesn't seem to be any support here for skipping this code if the default function hasn't been replaced? Did you not want to include that? It's just, if people don't replace the default config, this code will run unnecessarily as we know everything defaults to true.

lib/api/model.js
```diff
@@ -15,6 +17,101 @@ function get (modelName, passByReference) {
...
+ * @param {Function} cb        Callback function
+ * @return {Object}
+ */
+function getPermissions(req, modelName, cb) {
+
+    if (!linz.get('disabled permission cache') && !req.linz.cache.permissions.invalidate && req.session.permissions && req.session.permissions[modelName]) {
+        return cb(null, req.session.permissions[modelName]);
+    }
+
+    var model = get(modelName);
```
smebberson: I think you could safely retrieve a reference, rather than a copy here. We're not altering the model object so it should be fine, and no need for duplication...

lib/api/model.js
```diff
@@ -15,6 +17,101 @@ function get (modelName, passByReference) {
...
+        )
+        .filter(function (action) {
+            return Object.keys(action).indexOf('permission') >= 0;
+        })
+        .map(function (action) {
+            return action.permission;
+        })
+        .concat(['index','create','overview','edit','delete','json']);
+
+    async.seq(
```
smebberson: As with the `config.getPermissions` there is no support for default function detection to skip this code...

lib/api/configs.js
```diff
@@ -0,0 +1,80 @@
...
+        function (perms, callback) {
+
+            var permissionsObj = {};
+
+            permissions.forEach(function (permission) {
+                permissionsObj[permission] = perms.indexOf(permission) >= 0;
+            });
+
+            req.session.configs = req.session.configs || {};
+            req.session.configs = permissionsObj;
```
smebberson: Ah, maybe the error is here and it should be `req.session.configs[configName] = permissionsObj;`

lib/defaults.js
```diff
@@ -32,6 +32,14 @@ var defaults = module.exports = {
...
-
+	'set local time': false,
+	'permissions': function defaultPermissions (userId, action, context, callback) {
+		console.log('permissions');
+		return callback(null, true);
+	},
+
+	'disabled navigation cache' : false,
+	'disabled permission cache' : false,
+	'disabled config cache' : false
```
smebberson: These should all read _disable_ rather than _disabled_. They'll need to be changed everywhere else too.

lib/defaults.js
```diff
@@ -32,6 +32,14 @@ var defaults = module.exports = {
 	'cookie secret': 'oST1Thr2s/iOAUgeK4yuuA==',
 	'date format': 'ddd, ll',
 	'datetime format': 'llll',
-	'set local time': false
-
+	'set local time': false,
+	'permissions': function defaultPermissions (userId, action, context, callback) {
+		console.log('permissions');
```
smebberson: The `console.log` should be removed :)

lib/formtools/plugins/document.js
```diff
@@ -146,7 +146,7 @@ module.exports = function formtoolsPlugin (schema, options) {
 
     // defined get all data function
     // argument parameters is the same as http://mongoosejs.com/docs/api.html#model_Model.find
-    schema.statics.getQuery = function (filters, cb) {
+    schema.statics.getQuery = function (req, filters, cb) {
```
smebberson: How come `req` is passed in here, if it isn't used?

lib/router.js
```diff
@@ -20,15 +20,15 @@ exports.getRouter = function () {
...
 
     // map password reset
-    router.get(linz.get('admin password reset path') + '/:id/:hash', helpers.linz(), middleware.passwordReset.get, routes.passwordReset.get);
-    router.post(linz.get('admin password reset path'), helpers.linz(), middleware.passwordReset.post, routes.passwordReset.post);
+    router.get(linz.get('admin password reset path') + '/:id/:hash', middleware.passwordReset.get, routes.passwordReset.get);
+    router.post(linz.get('admin password reset path'), middleware.passwordReset.post, routes.passwordReset.post);
 
-	router.get('*', helpers.linz(), middleware.adminEnsureAuthenticated());
-	router.post('*', helpers.linz(), middleware.adminEnsureAuthenticated());
+	router.get('*', middleware.linz, middleware.navigation, middleware.adminEnsureAuthenticated());
```
smebberson: Was just reading this, not sure if possible, but it seems to make sense that the current order of `middleware.navigation` and then `middleware.adminEnsureAuthenticated()` would be swapped to `middleware.adminEnsureAuthenticated(),middleware.navigation`. Why generate the navigation unless they're authenticated?

lib/router.js
```diff
@@ -40,41 +40,40 @@ exports.setupModelRoutes = function () {
...
 	// model routes
 	router.get('/models/list', routes.modelList);
-	router.get('/model/:model/list', middleware.modelIndex.get, routes.modelIndex);
-    router.post('/model/:model/list', middleware.modelIndex.post, routes.modelIndex);
-	router.get('/model/:model/new', middleware.modelCreate(), routes.modelCreate);
-	router.post('/model/:model/create', middleware.modelSave(), routes.modelSave);
-    router.get('/model/:model/export', middleware.modelExport.get, routes.modelExport.get);
-    router.post('/model/:model/export', middleware.modelExport.post);
+	router.get('/model/:model/list', middleware.model, middleware.permissions('index', 'model'), middleware.modelIndex.get, routes.modelIndex);
+    router.post('/model/:model/list', middleware.model, middleware.permissions('index', 'model'), middleware.modelIndex.post, routes.modelIndex);
```
smebberson: I don't know why I didn't think of this before, but we could replace all of the `middleware.model` and `middleware.config` calls with Express's `app.param` functionality (http://expressjs.com/api.html#app.param). It would reduce the verbosity of our code.

lib/formtools/plugins/document.js
sandytrinh: This is required, as we need access to req.user.

lib/api/configs.js
sandytrinh: Fixed!

lib/api/model.js
sandytrinh: As discussed, we agreed there's no need to have a 'default' detection since we have caching in place so this bit of code will run once for every new session.

lib/api/configs.js
sandytrinh: Fixed!

lib/defaults.js
sandytrinh: Done!

lib/api/configs.js
sandytrinh: As discussed, we agreed there's no need to have a 'default' detection since we have caching in place so this bit of code will run once for every new session.

lib/api/model.js
sandytrinh: Done!

lib/defaults.js
sandytrinh: Done!

lib/router.js
sandytrinh: Done!

lib/router.js
sandytrinh: Replaced with `router.param`.

sandytrinh:
@smebberson I've implemented all changes. Please review!

sandytrinh:
@smebberson 

__Please note:__ I've updated linz permission parameters to simplify the data that is passed to the permission function. Here is the function signature:

```
function (user, action, context, callback){}

// example data for model permission check
action = 'index';
context = {
 type: 'model',
 name: 'rasUser'
};

// example data for config permission check
action = 'index';
context = {
 type: 'config',
 name: 'campaign'
};

// example data for navigation permission check
context = {
 type: 'navigation',
 name: 'Users',
 href: '/model/rasUsers',
 parent: {} // this field is available if applicable
};

```

lib/api/configs.js
```diff
@@ -0,0 +1,78 @@
...
+ * @param {String}   configName Config name
+ * @param {Function} cb        Callback function
+ * @return {Object}
+ */
+function getPermissions(req, configName, cb) {
+
+    if (!linz.get('disable config cache') && !req.linz.cache.permissions.configs.invalidate && req.session.configs && req.session.configs[configName]) {
+        return cb(null, req.session.configs[configName]);
+    }
+
```
smebberson: Could we please have some debug output here saying `Priming config cache.` The debug should use the `debug` module, with a setting of `linz:cache`. That will allow us to get a handle of when the cache is/isn't being regenerated.

lib/api/model.js
```diff
@@ -15,6 +17,101 @@ function get (modelName, passByReference) {
...
+ * @param {String}   modelName Model name
+ * @param {Function} cb        Callback function
+ * @return {Object}
+ */
+function getPermissions(req, modelName, cb) {
+
+    if (!linz.get('disable permission cache') && !req.linz.cache.permissions.models.invalidate && req.session.permissions && req.session.permissions[modelName]) {
+        return cb(null, req.session.permissions[modelName]);
+    }
+
```
smebberson: Could we please have some debug output here saying `Priming models cache.` The debug should use the `debug` module, with a setting of `linz:cache`. That will allow us to get a handle of when the cache is/isn't being regenerated.

lib/api/model.js
smebberson: Nice one, good call.

lib/api/navigation.js
```diff
@@ -0,0 +1,130 @@
+var linz = require('../../'),
+    utils = require('../utils'),
+    async = require('async');
+
+function get (req, cb) {
+
+    if (!linz.get('disable navigation cache') && !req.linz.cache.navigation.invalidate && req.session.navigation) {
+        return cb(null, req.session.navigation);
+    }
+
```
smebberson: Could we please have some debug output here saying `Priming navigation cache.` The debug should use the `debug` module, with a setting of `linz:cache`. That will allow us to get a handle of when the cache is/isn't being regenerated.

lib/formtools/plugins/document.js
smebberson: Okay, cool.

middleware-public/exclusions.js
```diff
@@ -0,0 +1,43 @@
+var async = require('async');
+
+module.exports = function () {
+
+    if (!arguments.length) {
+        throw new Error('Missing parameters! Please provide at least one middleware function.');
```
smebberson: I think this needs some context. Something like `Exclusions middleware is missing parameters. Please provide at least one middleware function.`

middleware/permissions.js
```diff
@@ -0,0 +1,60 @@
...
+
+};
+
+/**
+ * [getContextObj description]
+ * @param  {String} type The type of object for which permission is being requested.
+ * @param  {Object} req  The Express request object.
+ * @return {Object}      An object to be passed to the permissions functionality
+ */
+function getContextObj (type, req) {
```
smebberson: Is this no longer used? If so, it should be removed.

lib/api/configs.js
sandytrinh: Done!

lib/api/model.js
sandytrinh: Done!

lib/api/navigation.js
sandytrinh: Done!

middleware-public/exclusions.js
sandytrinh: Done!

middleware/permissions.js
sandytrinh: Done!

lib/api/navigation.js
smebberson: Upon running with this new build, I received the message 5 times. I think it should only be once?

lib/api/configs.js
```diff
@@ -0,0 +1,81 @@
...
+    debugCache('Priming configs cache');
```
smebberson: Looks good, but can it read:

```
debugCache('Priming config (' + configName + ') cache');
```

lib/api/model.js
```diff
@@ -15,6 +18,103 @@ function get (modelName, passByReference) {
...
+    debugCache('Priming models cache');
```
smebberson: Looks good, but can it read:

```
debugCache('Priming models (' + modelName + ') cache');
```

lib/api/navigation.js
sandytrinh: Unable to replicate this issue on your machine as well as mine.

lib/api/configs.js
sandytrinh: Done!

lib/api/model.js
sandytrinh: Done!

smebberson:
@sandytrinh, this is working well. Please merge!

PR #91, opened 2015-07-30
Use bootstrape-datepicker library

sandytrinh:
This PR changed the datepicker library to [bootstrape-datepicker](https://bootstrap-datepicker.readthedocs.org) as the current one we user [bootstrape-datetimepicker](https://eonasdan.github.io/bootstrap-datetimepicker/) as the datepicker doesn't work in a modal in IE browser.

However this new library has no support for time picker (whereas bootstrape-datetimepicker does)

There are few other issues that we haven't addressed for the datepicker in general:
- we don't have IE support for datetime widget
- the datepicker doesn't render the date in localised format
- the date post to the server must be in the format of 'yyyy-mm-dd' otherwise nodejs will throw an error.

_This is a working in progress PR!_


allanchau:
I don't think we need this anymore.

PR #92, opened 2015-08-04
Message template

sandytrinh:
This is a quick PR that adds a simple message jade template with similar layout to the login screen (see screenshot) that allows application to render a simple message.

image


smebberson:
@sandytrinh, this is all good. A nice addition too!

PR #93, opened 2015-08-14
Fixed issue where it only shows the first 20 records in model index.

sandytrinh:
Short PR.

@smebberson Please review!


smebberson:
@sandytrinh, Okay, let's merge, but should we come back and optimise this.

PR #95, opened 2015-12-09
Function based formtools, and permissions upgrade

smebberson:
This PR has two key parts:

- a re-implementation of a flexible and light weight permissions system for Linz.
- the ability to provide either an object, or a function for each of the keys within the model and config DSL

## Permissions

The following describes how permissions now works.

### Possible permissions and scope

Each permission, i.e. `canList` is scoped. For example `models.canList` and `model.canList`. The following breaks down the permissions, grouped by scope:

- models:
    - canList (can view a list of all models)
- model (i.e. a particular model such as rasUser):
    - canEdit (can edit any model record)
    - canDelete (can delete any model record)
    - canList (can view a list of all model records)
    - canCreate (can create any new model record)
    - canView (can view details of a particular model record)
-  configs:
    - canList (can view a list of all configs)
    - canView (can view a a particular config)
- config (i.e. a particular config such as googAnalytics):
    - canEdit (can edit any config record)
    - canReset (can reset any config)
    - canView (can view details of any config)

### Restriction implementation

Linz implements permission limits at two points:

- *In the UI:* it doesn't show buttons or options for functions that a user doesn't have access to.
- *Via the route:* routes are protected from one user (with permission) copying a route, and providing it to a user that doesn't have permission.

`models.canList` and `configs.canList` are the only two permissions routed through a global permissions function. All others are routed through models and configs themselves.

#### Global permissions function

Linz implementations can provide a function (in the `options` object when initialising Linz) called `permissions`. It should have the signature:

```
function permissions (user, context, permission, callback) {
}
```

The callback is a function with the following signature:

```
function callback (result) {
}
```

Result should be `true` or `false`. It does not accept an error. An error should be handled within implementation code, in case of an error, `false` should likely be returned.

This global function will be called whenever Linz is evaluating the `models.canList` and `configs.canList` permissions. This happens when evaluating navigation for a user, and the models list and configs list pages.

#### `model` and `config` permissions function

Determining permissions for models and configs is more context sensitive. To achieve this, when defining a model or config, you can also provide a `permissions` key. The key can have a value of either an `object`, or a `function`. If an object is provided, it will be used directly. If a function is provided it should have the following signature:

```
permissions: function (user, callback) {
}
```

Result should be `true` or `false`. It does not accept an error. An error should be handled within implementation code, in case of an error, `false` should likely be returned.

The callback is a function with the following signature:

```
function callback (result) {
}
```

The function provides the capability to have different permissions for different users. This is now the entire DSL for models:

```
collectionSchema.plugin(linz.formtools.plugins.document, {
    model: {
        ...
    },
    permissions:  {
        ...
    },
    grid: {
        ...
    },
    form: {
        ...
    },
    overview: {
        ...
    },
    fields: {
        ...
    }
});
```

### API

An API for permissions has been provided at `linz.api.permissions`. This interface makes it easy to work with Linz's permissions model. All internal code also uses this api.

## Todo

- [x] provide the ability to add to Linz's navigation structure (including a cache)
- [x] create a no-op function that will provide access to everything as a default
- [ ] exclude permissions code from executing, if the no-op function has been overwritten
- [x] implement the ability to control predefined permissions for **model routes**:
    - [x] model create
    - [x] model edit
    - [x] model delete
    - [x] model json
    - [x] model export
    - [x] model overview
- [x] implement the ability to control **model route** permissions on custom model actions
- [x] implement the ability to control predefined permissions for **model views**:
    - [x] model create
    - [x] model edit
    - [x] model delete
    - [x] model json
    - [x] model export
    - [x] model overview
- [x] implement the ability to control **model view** permissions for custom model actions
- [x] implement the ability to control predefined permissions for **config routes**:
    - [x] config overview
    - [x] config reset
    - [x] config edit
    - [x] config json
- [x] implement the ability to control predefined permissions for **config views**:
    - [x] config overview
    - [x] config reset
    - [x] config edit
    - [x] config json
- [ ] cache data per user session for:
 - [ ] navigation
 - [ ] model permissions
 - [ ] configs permissions
- [x] provide a public middleware `linz.middleware.exclusions`
- [x] update model DSL to accept either an `object` or a `function`
 - [x] update `model.grid`
 - ~~update `model.model`~~
 - [x] update `model.form`
 - [x] update `model.overview`
 - ~~update `model.fields`~~
- [x] update config DSL to accept either an `object` or a `function`
 - [x] update `config.grid`
 - [x] update `config.form`
 - [x] update `config.overview`
 - [x] update `config.fields`
- [x] Implement a labels structure
 - [x] Use labels object in the form
 - [x] Override values from the labels object in the form
 - [x] Use labels object in the overview
 - [x] Override values from the labels object in the overview
 - [x] Use labels object in the grid
 - [x] Override values from the labels object in the grid
 - [x] Use labels object in filters
 - [x] Override values from the labels object in filters
 - [x] Use labels object in sortBy


sandytrinh:
@smebberson I've tided up a few things in this PR:

- [x] Updated `document.js` and `embeddedDocument.js` to load the `options` default on application start is `options.grid`, `options.form` and `options.overview` are not provided as function. This will greatly optimise the call as each functions runs only ones rather on each function call.
- [x] Updated model.labels so any labels that are not provided will automatically default to the field name
- [x] Fixed the overview `canEdit` and `canDelete` function for each record
- [x] Fixed up export functionality where `getGrid` wasn't refactored to include `user` data and I also removed any references to `modelName` as we have access to the model from `req.linz.model`.

Questions:

- Why do we need to provide a permission function to block the config routes? How is this handled for model?

smebberson:
@sandytrinh. Nice updates. Thanks for that. Should we merge?

lib/formtools/plugins/plugins-helpers.js
```diff
@@ -390,6 +383,12 @@ var formSettings = function (schema, form, labels) {
     for (var key in formConfigs){
         schemaConfig = formConfigs[key] || {};
         formConfig = form[key] || {};
+
+        // support for form DSL where "fieldName: true" with no other configurations
+        if (typeof formConfig === 'boolean' && formConfig === true) {
```
smebberson: You could remove the `typeof formConfig === 'boolean' &&` part, tripe equals will suffice.

lib/formtools/plugins/document.js
```diff
@@ -57,6 +75,19 @@ module.exports = function formtoolsPlugin (schema, options) {
         });
     }
 
+    // check if we need to render the default if grid, form and overview options are provided in object DSL
+    if (typeof options.grid !== 'function') {
+        var grid = pluginHelpers.gridDefaults(opts.grid, labels);
```
smebberson: With this and the next two, the variables should be locally scoped above on L18.

lib/formtools/plugins/plugins-helpers.js
sandytrinh: Done!

lib/formtools/plugins/document.js
sandytrinh: Done!

PR #97, opened 2016-02-03
Redirect after login

smebberson:
This PR will redirect the user after successful login to the original request url they were attempting to resolve if the request is hijacked by the authorisation middleware.

To recreate:

- make sure you're logged out of Linz
- enter a protected URL in your browser (not `/admin/models/list` as that is the default)
- you'll be redirected to the login screen
- login to Linz
- after login, you should be on the protected URL you original entered rather than `/admin/models/list`

Further testing:

- [x] user should be redirected to original URL if they attempted to resolve a protected URL while logged out
- [x] should not break current functionality, such as redirecting to the default Linz admin url of `/admin/models/list`

To test this against a plain Linz setup, use the recently updated repository https://github.com/smebberson/linz-minitwitter-basic. It comes with a Vagrant setup for easy installation.

### DIFF
--- a/lib/api/url.js
+++ b/lib/api/url.js
@@ -56,7 +56,7 @@ exports.getAdminLink = function (obj, action, id) {
 
 };
 
-exports.getLink = function (links) {
+exports.getLink = function () {
 
     var url = linz.get('admin path');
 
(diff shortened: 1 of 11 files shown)


smebberson:

sandytrinh:
@smebberson Awesome new feature! Works really well!
Please merge!

PR #102, opened 2016-05-27
Upgrade overview UI

patelswapnil:
### This PR improves the overview UI and allows the user to define the Overview as follow:
- Render `Date created` and `Date modified` as default Summary when `overview` or `overview.body` is not set in overview configuration
- User can group fields to be rendered under given title in Overview
- User can pass their own function to render the data as they wish 
- User can group fields to be rendered under given title in seperate tabs
- User can pass their own function to render data as they wish under each tab
### Usage
- A function or an object can be provided to `overview`
- Provide custom function to overview
  **Note:** This function is passed a `user` and a `callback`. A callback must be returned with object.

```
e.g. 1:
overview: function (user, callback) {

        var overviewBody = {
            body: function (req, res, record, model, cb) {
                    return cb(null, '<p>Content Printed by function set to overview instead of object<p>');
            }
        };

        return callback(null, overviewBody);

    }

e.g. 2:
overview: function (user, callback) {

    var overviewBody = {
        body: [
            {
                label: 'label 1',
                fields: ['field1', 'field2', ...]
            },
            {
                label: 'label 2',
                fields: ['field3', 'field4', ...]
            },
            .... more field groups ....
        ]
    };

    return callback(null, overviewBody);

}
```
- Provide an object to `overview`: following variations can be set to `overview` object
  - set custom function to overview.body to render data as required
  
  ```
  overview: {
       body: function (req, res, record, model, callback) {
            return callback(null, '<p>Custom overview body data<p>');  
       }
  }
  ```
  - Group fields to render them under one title:
  
  ```
  overview: {
    body: [
      {
        label: 'label 1',
        fields: ['field1', 'field2', ...]
      },
      {
        label: 'label 2',
        fields: ['field3', 'field4', ...]
      },
      .... more field groups ....
    ]
  }

  ```

  - Note: fields key in object in overview.body accepts array. You can provide a field names and objects to these array to render those fields under one label as filed group:

       - field name as string
       - object with `fieldName` and `label` to provide custom label for filed
       - object with `fieldName` and `renderer` to render the filed value by one of the cell renderers provided by linz
           - e.g use linz.formtools.cellRenderers.url to render the field value as url.
       - object with `label` and `custom function` to process complex values and display in desired format under custom label

        e.g:

        fields: [
             'field1', 
             'field2',
             {
                 fieldName: 'field3',
                 label: 'field label'
             },
             {
                 fieldName: 'field4',
                 renderer: linz.formTools.cellRenderers.date
             },
             {
                 label: 'field label',
                 value: function(req, res, record, model, callback) {
                         return <html content>;
                 }
             }
        ]

  - Provide your custom function to render the data as you like:
  
  ```
  overview: {
    body: [
      {
        label: 'Custom Data',
        body: function (req, res, record, model, callback) {
          .... function code to create html content here....
          return callback(null, 'html content here');
        }
      },
      .... more field groups ....
    ]
  }
  ```
  
     **Note:** Linz will provide four arguments (**_req_**, **_res_**, **_record_**, **_model_**) to custom function in overview.
     Your custom function should return **html content**.
  - Fields can be grouped and rendered under tabs as follow:
  
  ```
  overview: {
    body: [
      [
        {
          label: 'Tab 1',
          fields: ['field1', 'field2', ...]
        },
        {
          label: 'Tab 2',
          fields: ['field3', 'field4', ...]
        },
        .... more tabs ....
      ]
    ]
  }
  ```
  - Provide your custom function to render data as you like under tab
  
  ```
  overview: {
    body: [
      [
        {
          label: 'Custom Tab',
          body: function (req, res, record, model, callback) {
            .... function code to create html content here....
            return callback(null, 'html content here');
          }
        },
        .... more tabs ....
      ]
    ]
  }
  ```
  - A detailed example of how overview can be configured:
  
  ```
  overview: {
    body: [
      // fieldsets (grouping fields under one title)
      {
        label: 'Label 1',
        fields: ['field1', 'field2', ...]
      },
      {
        label: 'Label 2',
        body: function (req, res, record, model, callback) {
          .... function code to create custom html content here ....
          return callback(null, 'html content here');
        }
      },
      // .... more fieldsets .... 
      // tabs
      [
        {
          label: 'Tab 1',
          fields: ['field1', 'field2', ...]
        },
        {
          label: 'Tab 2',
          body: function (req, res, record, model, callback) {
            .... function code to create custom html content here ....
            return callback(null, 'html content here');
          }
        },
        // example of tab containing multiple fieldsets
        {
          label: 'Tab containing multiple fieldsets',
          body: [
            {
              label: 'label 1',
              fields: ['field1', 'field2', ...]
            },
            {
              label: 'Label 2',
              body: function (req, res, record, model, callback) {
                .... function code to create custom html content here....
                return callback(null, 'html content here');
              }
            },
            .... more fieldsets ....
          ]
        },
        {
          label: 'Tab 4',
          fields: ['field3', 'field4', ...]
        }
        .... more tabs ...
      ], // end of tabs
      // few more fieldsets if neded after tabs 
      {
        label: 'Label 3',
        fields: ['field3', 'field4', ...]
      },
      {
        label: 'Label 4',
        body: function (req, res, record, model, callback) {
          .... function code to create custom html content here ....
          return callback(null, 'html content here');
        }
      },
      // .... more fieldsets ....
      // .... more tabs ....
    ] // end of overview.body
  } // end of overview
  ```


patelswapnil:
@smebberson and @sandytrinh this PR is ready for review

lib/formtools/plugins/plugins-helpers.js
```diff
@@ -226,6 +226,41 @@ var overviewDefaults = function (overview, labels) {
...
+        overview.details = {};
+    }
+
+    overview.details.form = overviewDetailsDefault(overview.details.fields, schema, form, labels);
+
+    //set tabs to tabsArray property of overview.details
+    var tabsArray = [];
+    if (Array.isArray(overview.details.tabs) && overview.details.tabs.length) {
+
+        for (var i = 0; i < overview.details.tabs.length; i++) {
```
smebberson: Can this `for` loop please be replaced with a `forEach`? The syntax is nicer and cleaner.

lib/formtools/plugins/plugins-helpers.js
```diff
@@ -226,6 +226,41 @@ var overviewDefaults = function (overview, labels) {
...
+
+            var tempTab = {};
```
smebberson: This could be combined:

```
var tempTab = {
    label: overview.details.tabs[i].label;
};
```

lib/formtools/plugins/plugins-helpers.js
```diff
@@ -226,6 +226,41 @@ var overviewDefaults = function (overview, labels) {
 
     overview.summary.label = overview.summary.label || 'Summary';
 
+
+    //overview Details
+    if (!overview.details || typeof overview.details !== 'object') {
+        // defaults to {}
```
smebberson: This comment isn't helpful. It can be removed. I prefer to keep comments to provide information that isn't obvious from reading the code. Could you please also ensure comments are in the format `// comment here` (noting the space between `//` and the first part of the comment).

lib/formtools/plugins/plugins-helpers.js
```diff
@@ -226,6 +226,41 @@ var overviewDefaults = function (overview, labels) {
...
+        overview.details = {};
+    }
+
+    overview.details.form = overviewDetailsDefault(overview.details.fields, schema, form, labels);
+
+    //set tabs to tabsArray property of overview.details
+    var tabsArray = [];
```
smebberson: Could you please rename `tabsArray` to `tabs` and create it via `.map` on a clone of `overview.details.tabs`? For example:

```
var tabs = clone(overview.details.tabs).map(function (tab) {

    // code here...

});
```

lib/formtools/plugins/plugins-helpers.js
```diff
@@ -226,6 +226,41 @@ var overviewDefaults = function (overview, labels) {
...
+            } else if (typeof overview.details.tabs[i].body === 'function') {
+                tempTab.body = overview.details.tabs[i].body;
+            }
+
+            tabsArray.push(tempTab);
+        }
+    }
+
+    //set the tabsArray to overview.details.tabsArray
+    overview.details.tabsArray = tabsArray;
```
smebberson: Could you please rename `overview.details.tabsArray` to be `overview.details.tabs`. This will keep it more consistent with the rest of Linz. We haven't been using data type based naming anywhere else.

routes/recordOverview.js
```diff
@@ -31,6 +34,138 @@ var route = function (req, res, next) {
...
+        //process tabs of overview detials
+        function (cb) {
+
+            if (!Array.isArray(req.linz.model.linz.formtools.overview.details.tabsArray) || !req.linz.model.linz.formtools.overview.details.tabsArray.length) {
+                return cb(null);
+            }
+
+            (function processTabs(i) {
+
+                var detailTab = req.linz.model.linz.formtools.overview.details.tabsArray[i];
```
smebberson: Linz styling is more like:

```
var detailTab = req.linz.model.linz.formtools.overview.details.tabsArray[i],
    tempTab = {
        label: detailTab.label
    };
```

routes/recordOverview.js
```diff
@@ -31,6 +34,138 @@ var route = function (req, res, next) {
...
+                var tempTab = {};
+                tempTab.label = detailTab.label;
+
+                if (typeof detailTab.form === 'object') {
```
smebberson: I'm not a big fan of the `if`, `else if`, `else` pattern. Could this please be rewritten. It's not used much in Linz.js anywhere. If you peruse the source code, you should find some better alternatives (such as `return` early).

routes/recordOverview.js
```diff
@@ -31,6 +34,138 @@ var route = function (req, res, next) {
...
+
+                    //make sure detailTab.form is not an empty object
+                    if (utils.isEmptyObject(detailTab.form)) {
+
+                        i++;
```
smebberson: What is this for?

routes/recordOverview.js
```diff
@@ -31,6 +34,138 @@ var route = function (req, res, next) {
...
+        },
+
+        //process tabs of overview detials
+        function (cb) {
+
+            if (!Array.isArray(req.linz.model.linz.formtools.overview.details.tabsArray) || !req.linz.model.linz.formtools.overview.details.tabsArray.length) {
+                return cb(null);
+            }
+
+            (function processTabs(i) {
```
smebberson: Could you please explain the benefit of this self calling function iterator thingo over a standard `forEach` or `map`?

routes/recordOverview.js
```diff
@@ -31,6 +34,138 @@ var route = function (req, res, next) {
...
+            })(0);
+
+        },
+
+        //set tabs content to overviewDetail
+        function (cb) {
+
+            //render locals.tabs set by previous function in async.series
+            if (!locals.tabs.length) {
+
```
smebberson: This empty line could be removed.

smebberson:
I'm not sure keen on this using the edit UI. I don't like the fact that it renders everything as disabled forms, or that it changes the format of the UI away from what is shown in the summary.

Regarding the UI moving away from what is shown in the summary, I don't mind if that moves closer to the new format (i.e. right-aligned, etc) but it needs to be consistent.

I love the tabs and the overall approach to this, but I think the rendering of each field should not be based on form elements. We need to recreate the formtool renderers that are focussed on rendering data outside the context of the form (i.e. just static text). I think this will be useful in many more scenarios that just this too. Very reusable.

I know this is more work, but it's just not sharp enough at the moment.

lib/formtools/render-overview.js
```diff
@@ -0,0 +1,88 @@
...
+        default:
+            cellrenderer = 'default';
+            break;
+
+    }
+
+    return cellrenderer;
+};
+
+var generateViewFromModel = exports.generateViewFromModel = function (schema, schemaFields, record, model, callback) {
```
sandytrinh: A better name for this function is perhaps `render`? And you should structure it like the example below:

```
function render () {};
module.exports = {
  render: render
}
```

lib/formtools/render-overview.js
```diff
@@ -0,0 +1,88 @@
...
+    var fieldsets = {};
+
+    async.eachSeries(Object.keys(schemaFields), function (fieldName, cb) {
+
+        var field = schemaFields[fieldName],
+            fieldDetail = {
+                label: field.label || ''
+            };
+
+        var title = field.fieldset || 'Summary';
```
sandytrinh: This bit of logic is not correct. What you are programmed here is if a fieldset is not provided, it will default to summary. That means any fields that doesn't have a fieldset will be grouped into this 'Summary' fieldset. According to how the edit handler works is any fields with no fieldset will be add to the bottom of the form with no grouping. The overview so behave the same as the form edit handler.

lib/formtools/render-overview.js
```diff
@@ -0,0 +1,88 @@
...
+        if (!fieldsets[title]) {
+            fieldsets[title] = [];
+        }
+
+        var cellRenderer = (cellRenderers[field.type]) ? field.type : mapCellRenderer(field.type);
```
sandytrinh: Rename this to `renderer` so there's a clear distinction on `L69`.

lib/formtools/render-overview.js
```diff
@@ -0,0 +1,88 @@
...
+
+        cellRenderers[cellRenderer](record[fieldName], record, fieldName, model, function (err, value) {
+
+            if (err) {
+                console.log('Encountered error while rendering ' + field.type + ' via cellRenderers from render-overview.js for the field ' + fieldName);
+                console.log(err);
```
sandytrinh: Make sure you remove these `console.log`

lib/formtools/render-overview.js
```diff
@@ -0,0 +1,88 @@
...
+                return cb(err);
+            }
+
+            fieldDetail.value = value;
+            fieldsets[title].push(fieldDetail);
```
sandytrinh: Since `fieldDetail` variable is used only once, you will simplify to the code below:

```
fieldsets[title].push({
  label: field.label
});
```

Note that I removed `|| ''` from the label, because label will always exits. If a label is not defined, it will default to the fieldName. Form config has handled this for you.

lib/utils.js
```diff
@@ -137,3 +137,20 @@ exports.noOp = function (_cb) {
 	}
 
 };
+
+/**
+ * Return true if object is empty else false
+ * @param {Object} obj   Object to be tested if it is empty or not
+ * @return {Bollean}
+ */
+exports.isEmptyObject = function (obj) {
```
sandytrinh: Not sure why you need this function, we normally use `Object.keys(obj).length` to determine is an object has any keys.

routes/configOverview.js
```diff
@@ -30,6 +30,28 @@ var route = function (req, res) {
 
         },
 
+        function (cb) {
```
sandytrinh: Make sure you test that this works for config as well. Config data is a bit different from models.

routes/recordOverview.js
```diff
@@ -1,7 +1,35 @@
...
 var linz = require('linz'),
     async = require('async'),
-    clone = require('clone');
+    clone = require('clone'),
+    utils = require('../lib/utils');
+
+
+var filterForm = function (form, formFields) {
+
+    if (typeof form !== 'object' || utils.isEmptyObject(form)) {
```
sandytrinh: You can use `!Object.keys(form).length` to determine the object has no keys.

routes/recordOverview.js
```diff
@@ -1,7 +1,35 @@
...
+        return {};
+    }
+
+    if (!Array.isArray(formFields) || !formFields.length) {
```
sandytrinh: You can combine `L10` and `L14` in one if statement.

routes/recordOverview.js
```diff
@@ -10,27 +38,131 @@ var route = function (req, res, next) {
...
 
-            req.linz.model.linz.formtools.overview.summary.renderer(req.linz.record, req.linz.model, function (err, content) {
+            var details = req.linz.model.linz.formtools.overview.details;
 
-                if (err) {
-                    return cb(err);
-                }
+            //render default overview if details is not provided or details provided is not a function OR not an Array containing fileds
+            if ( !details || !(typeof details === 'function' || (Array.isArray(details) && details.length)) ) {
+
```
sandytrinh: Shouldnn't `details` always exists? If user hasn't defined `overview.details` linz should default it to `['dateCreated','dateModified']`

routes/recordOverview.js
```diff
@@ -10,27 +38,131 @@ var route = function (req, res, next) {
...
+            if (!Array.isArray(req.linz.model.linz.formtools.overview.tabs) || !req.linz.model.linz.formtools.overview.tabs.length) {
 
                 return cb(null);
+            }
 
-            });
+            async.eachSeries(req.linz.model.linz.formtools.overview.tabs, function (tab, callback) {
+
+                if ( !tab.fields && !tab.body ) {
+                    return callback(null);
```
sandytrinh: You can omit `null` in the callback function.

views/overview/tabs.jade
```diff
@@ -0,0 +1,20 @@
...
+        div
+            ul.nav.nav-tabs.nav-tabs-overview
+                each tab, index in tabs
+                    li(class="#{index === 0 ? 'active' : ''}")
+                        a(href='#' + tab.label + index, role='tab' data-toggle='tab')
+                            span=tab.label
+        div.tab-content
+            each tab, index in tabs
+                if (tab.fields)
+                    div(id=tab.label + index, class="#{index === 0 ? 'tab-pane active' : 'tab-pane'}")
```
sandytrinh: Should move this above the if statement on `L13` and you can remove the duplicates from the rest of the `if` statements since you need to output that markup regardless of the `if` statements.

patelswapnil: done.

routes/recordOverview.js
patelswapnil: done.

routes/recordOverview.js
patelswapnil: done.

routes/recordOverview.js
patelswapnil: done.

routes/recordOverview.js
patelswapnil: done.

routes/configOverview.js
patelswapnil: done.

lib/utils.js
patelswapnil: removed the `isEmptyObject` function that was added to utils.js and all its references used in overView UI code. replaced it  with `Object.keys(<obj>).length` to determine if object has any keys.

lib/formtools/render-overview.js
patelswapnil: done.

lib/formtools/render-overview.js
patelswapnil: done.

lib/formtools/render-overview.js
patelswapnil: done.

lib/formtools/render-overview.js
patelswapnil: done.

lib/formtools/render-overview.js
patelswapnil: done.

routes/recordOverview.js
patelswapnil: done.

routes/recordOverview.js
patelswapnil: replaced the self executing function with async.eachSeries to loop through tabs and to execute callback on each tab asynchronously.

routes/recordOverview.js
patelswapnil: done.

routes/recordOverview.js
patelswapnil: done.

lib/formtools/plugins/plugins-helpers.js
patelswapnil: done.

lib/formtools/plugins/plugins-helpers.js
patelswapnil: done. Code to process overview.details has been removed from plugins-helpers.js. It has been moved to formtools.overview

lib/formtools/plugins/plugins-helpers.js
patelswapnil: done. Code to process overview.details has been removed from plugins-helpers.js. It has been moved to formtools.overview.

lib/formtools/plugins/plugins-helpers.js
patelswapnil: done. Code to process overview.details has been removed from plugins-helpers.js. It has been moved to formtools.overview.

lib/formtools/plugins/plugins-helpers.js
patelswapnil: done. Code to process overview.details has been removed from plugins-helpers.js. It has been moved to formtools.overview.

patelswapnil:
@sandytrinh, all suggestions have been implemented and committed. You may review.

lib/formtools/render-overview.js
```diff
@@ -0,0 +1,89 @@
...
+            cellrenderer = 'default';
+            break;
+
+    }
+
+    return cellrenderer;
+};
+
+function render(schema, schemaFields, record, model, callback) {
+
```
sandytrinh: Perhaps renamed `schemaFields` to `fields` so that code is easier to follow.

lib/formtools/render-overview.js
```diff
@@ -0,0 +1,89 @@
...
+    async.eachSeries(Object.keys(schemaFields), function (fieldName, cb) {
+
+        var field = schemaFields[fieldName],
+            title = field.fieldset || '(No fieldset defined)';
+
+        if (!fieldsets[title]) {
+            fieldsets[title] = [];
+        }
+
```
sandytrinh: You can combine `L60` to `L62` with this one-liner:

```
fieldsets[title] = fieldsets[title] || []
```

views/overview/tabs.jade
```diff
@@ -0,0 +1,17 @@
...
+        div
+            ul.nav.nav-tabs.nav-tabs-overview
+                each tab, index in tabs
+                    li(class="#{index === 0 ? 'active' : ''}")
+                        a(href='#' + tab.label + index, role='tab' data-toggle='tab')
+                            span=tab.label
+        div.tab-content
+            each tab, index in tabs
+                div(id=tab.label + index, class="#{index === 0 ? 'tab-pane active' : 'tab-pane'}")
+                    if (tab.fields)
```
sandytrinh: This should check `if (tab.body)` first so it's consistent with your code logic in overview.render.

lib/formtools/render-overview.js
patelswapnil: done.

lib/formtools/render-overview.js
patelswapnil: done.

views/overview/tabs.jade
patelswapnil: done.

patelswapnil:
@sandytrinh the PR has been updated. You may review it.

views/mixins/overviewFields.jade
```diff
@@ -0,0 +1,10 @@
+mixin overviewDetails(fields)
+	each fieldset, key in fields
+		section.group
+			h1=key
+			div.body
+				each field in fieldset
+					div.field
+						p
+							span.col-sm-2.label=field.label
```
sandytrinh: These 2 lines show be wrapped in a `row` class. See https://getbootstrap.com/css/#grid for more info on the structure of the grid. Perhaps the class `overview-details` should also have a class named `container`

views/configOverview.jade
```diff
@@ -35,8 +35,8 @@ block content
 	.container.linz-container.overview.config
 		.pane
 			.body
-				.overview-summary
-					div!= overviewSummary
+				.overview-detail
+					include overview/details
```
sandytrinh: Please rename this to `.overview-details` for consistency.

views/recordOverview.jade
```diff
@@ -50,8 +50,10 @@ block content
 	.container.linz-container.overview.config
 		.pane
 			.body
-				.overview-summary
-					div!= overviewSummary
+				.overview-detail
```
sandytrinh: Please rename this to `.overview-details` for consistency.

views/configOverview.jade
patelswapnil: done.

views/recordOverview.jade
patelswapnil: done.

views/mixins/overviewFields.jade
patelswapnil: add row class to div that wraps fields. container class is already added to the outer most div.

lib/formtools/plugins/plugins-helpers.js
```diff
@@ -23,8 +23,8 @@ var defaults = function (options, isEmbeddedDocument) {
     }
 
     // dateCreated and dateModified will be hidden by default
```
smebberson: Should this comment read **shown** rather than hidden?

patelswapnil: In the form dateCreated and dateModified fields are set to be hidden by default. So, I think the comment is correct that these two fields will be hidden by default in form.

lib/formtools/overview.js
```diff
@@ -0,0 +1,56 @@
+
+var async = require('async'),
+    cellRenderers = require('./renderers-cell');
+
```
smebberson: Is there one too many empty lines here?

lib/formtools/overview.js
```diff
@@ -0,0 +1,56 @@
...
+
+function getOverviewFields(schema, form, selectedFields, record, model, callback) {
+
+    var fieldsets = {},
+        fields = {};
+
+    selectedFields.forEach( function(fieldName) {
```
smebberson: This code would be more performant if written:

```
var ref = form[fieldName];

if (ref) {
    fields[fieldName] = {
        label: ref.label,
        type: ref.type,
        fieldset: ref.fieldset
    }
}
```

lib/formtools/overview.js
```diff
@@ -0,0 +1,56 @@
...
+    });
+
+    async.eachSeries(Object.keys(fields), function (fieldName, cb) {
+
+        var field = fields[fieldName],
+            title = field.fieldset || '(No fieldset defined)';
+
+        fieldsets[title] = fieldsets[title] || [];
+
+        var renderer = (cellRenderers[field.type]) ? field.type : 'text';
```
smebberson: Can this `var` statement be grouped with the others above?

lib/formtools/overview.js
```diff
@@ -0,0 +1,56 @@
...
+            if (err) {
+                return cb(err);
+            }
+
+            fieldsets[title].push({
+                label: field.label,
+                value: value
+            });
+
+            return cb();
```
smebberson: Inside a function that has more than one line of code, usually the first line has an empty line before it, and the last line has an empty line after it.

lib/formtools/overview.js
```diff
@@ -0,0 +1,56 @@
...
+        });
+
+    }, function (err) {
```
smebberson: Inside a function with only one line of code, we remove all empty lines. Sorry, confusing!

lib/formtools/renderers-cell.js
```diff
@@ -156,7 +156,16 @@ var documentArrayRenderer = module.exports.documentArray = function documentArra
 
     });
 
-}
+};
+
+var textRenderer = module.exports.text = function textRenderer(val, record, fieldName, model, callback) {
+
+    if (val === undefined) {
```
smebberson: The best way to check for undefined in this instance is to use `typeof val === 'undefined'` for reasons describe in this thread https://stackoverflow.com/questions/4725603/variable-undefined-vs-typeof-variable-undefined

routes/recordOverview.js
```diff
@@ -3,34 +3,101 @@ var linz = require('linz'),
...
+                return cb();
             });
 
         },
 
+        // process overview tabs
+        function (cb) {
+
+            if (!Array.isArray(req.linz.model.linz.formtools.overview.tabs) || !req.linz.model.linz.formtools.overview.tabs.length) {
+
```
smebberson: You don't need this extra line...

routes/recordOverview.js
```diff
@@ -3,34 +3,101 @@ var linz = require('linz'),
...
+                            label: tab.label,
+                            body: content
+                        });
+
+                        return callback();
+                    });
+
+                }
+
+                if ( !Array.isArray(tab.fields) || !tab.fields.length ) {
```
smebberson: Please remove the preceding and post spaces within the brackets (i.e. `if (!Array.isArray(tab.fields) || !tab.fields.length) {`).

lib/formtools/overview.js
patelswapnil: done.

lib/formtools/overview.js
patelswapnil: done.

lib/formtools/overview.js
patelswapnil: done.

lib/formtools/overview.js
patelswapnil: done.

lib/formtools/overview.js
patelswapnil: done.

lib/formtools/renderers-cell.js
patelswapnil: done.

routes/recordOverview.js
patelswapnil: done.

routes/recordOverview.js
patelswapnil: done.

smebberson:
@patelswapnil, Linz tries to provide a good UI on mobile where it can. This can get tricky at times. However, for this instance, the mobile UI needs a little work. It doesn't seem to render properly on a mobile viewport.

smebberson:
@patelswapnil, can you please implement the following styling updates?

- Please remove the grey background for field values (and readjust the alignment of the text so it feels natural).
- Please remove the border around the tabs such that only the top border exists. You can then remove the padding around the content within the tabs.

Let's start with those two.

smebberson:
@patelswapnil, also, can you please verify that this doesn't break this earlier PR (https://github.com/linzjs/linz/pull/95) in which `overview` can be either an `object` or a `function`?

smebberson:
@patelswapnil, @sandytrinh while I like the functionality, I'm wondering if we can extend the DSL to something like:

```
overview: {
    body: [
        {
            label: 'Summary',
            fields: ['businessName', 'code', 'type', 'bActive']
        },
        [
            {
                label: 'Details',
                fields: ['slug', 'bHasConsumerWebsite', 'address2', 'suburb', 'state', 'postcode']
            },
            {
                label: 'Services',
                fields: ['specialistServices']
            },
            {
                label: 'Test',
                body: function (req, res, record, model, callback) {
                    return callback(null, '<p>Custom render under the test tab.</p>');
                }
            }
        ],
        {
            label: 'Another fieldset below',
            fields: ['something', 'somethingelse']
        }
    ]
}
```

`overview.body` would accept a `function` that returns an `array` or an `array` itself.

While looping over the `body` `array`, if an `object` is encountered it would render a field set, if an `array` is encountered it would render a set of tabs.

I think this would provide us with the ultimate flexibility to quickly and easily create worthwhile and useful overview pages.

What do you both think?

sandytrinh:
@smebberson I like your suggestion, it that DSL, it is quite flexible but I have a few question below:

Currently with `overview.details` a user can simply specify the fields to display, and Linz will render them group by fieldsets as per the `form` DSL. If we move to your suggestion, that means a user will have to specify each fieldset and properties to render. This moves away from our original goal - which was to effortlessly render the content in the overview (that resembles the edit handler) with minimum configurations since we can use the DSL from `form`.

smebberson:
@sandytrinh, I agree it is definitely a shift. However, if you take a look at some of the implementations we have (i.e. private projects) you'll note they're not easy to achieve given this new setup. While I think this suggested DSL is a _little_ more work, it's **much** easier to achieve more complex layouts with this newly suggested DSL than the currently implemented one.

```
overview: {
    body: [
        label: 'Summary',
        fields: 'one, two, three, four'
    ]
}
```

is only a little more typing than:

```
overview: {
    details: ['fieldName1', 'fieldName2']
}
```

I think it's a good trade off between ease of use at both the simple level, and the sophisticated level.

With the newly suggested DSL it's easy to have a layout such as:
```
- Fieldset
    - Field 1
    - Field 2
    - Field 3
- Tabs
    - Tab 1
        - Fieldset
            - Field 1
            - Field 2
    - Tab 2
        - Field 1
        - Field 2
- Fieldset
    - Field 1
    - Field 2
- Fieldset
    - Field 1
```

Whereas anything under the tabs is hard to achieve at the moment.

sandytrinh:
Let's the take the follow form DSL as an example:

```
form: {
    firstname: {
        fieldset: 'Member details'
    },
    lastName: {
        fieldset: 'Member details'
    }
    email: {
        fieldset: 'Member details'
    },
    street: {
        fieldset: 'Member location'
    },
    suburb: {
        fieldset: 'Member location'
    },
    postcode: {
        fieldset: 'Member location'
    },
    country: {
        fieldset: 'Member location'
    },
    street: {
        fieldset: 'Member location'
    },
    openMonday: {
        fieldset: 'Opening hours'
    },
    closeMonday: {
        fieldset: 'Opening hours'
    },
    openSaturday: {
        fieldset: 'Opening hours'
    },
    closeSaturday: {
        fieldset: 'Opening hours'
    },
    openSunday: {
        fieldset: 'Opening hours'
    },
    closeSunday: {
        fieldset: 'Opening hours'
    }
}
```

If we want to have it render as per the edit handler, the `overview` DSL would look like below:

```
overview: {
    body: [
        {
            label: 'Member details',
            fields: ['firstName', 'lastName', 'emails']
        },
        {
            label: 'Member location',
            fields: ['street', 'suburb', 'postcode', 'country']
        },
        {
            label: 'Opening hours',
            fields: ['openMonday', 'closeMonday', 'openSaturday', 'closeSaturday', 'openSunday', 'closeSunday']
        }
    ]
}
```

If we want to break it into 3 tabs instead, the `overview` DSL would look like so:

```
overview: {
    body: [
        [
            {
                label: 'Member details',
                fields: ['firstName', 'lastName', 'emails']
            },
            {
                label: 'Member location',
                fields: ['street', 'suburb', 'postcode', 'country']
            },
            {
                label: 'Opening hours',
                fields: ['openMonday', 'closeMonday', 'openSaturday', 'closeSaturday', 'openSunday', 'closeSunday']
            }
        ]
    ]
}
```

What would the `overview`  DSL look like, if we need to have multiple fieldset within a tab?

smebberson:
```
overview: {
    body: [
        [
            {
                label: 'Member details',
                fields: ['firstName', 'lastName', 'emails']
            },
            {
                label: 'Member location',
                body: [
                    {
                        label: 'Fieldset 1',
                        fields: ['street', 'suburb', 'postcode', 'country']
                    },
                    {
                        label: 'Fieldset 2',
                        fields: ['street', 'suburb', 'postcode', 'country']
                    }
                ]
            },
            {
                label: 'Opening hours',
                fields: ['openMonday', 'closeMonday', 'openSaturday', 'closeSaturday', 'openSunday', 'closeSunday']
            }
        ]
    ]
}
```

sandytrinh:
So ultimately the `overview` DSL would except the following:

```
overview: {
    body: [
        {
            label: 'Summary',
            fields: ['businessName', 'code', 'type', 'bActive']
        },
        {
            label: 'Graphs',
            body: function (req, res, record, model, callback) {
                return callback(null, '<p>Big pretty graph.</p>');
            }
        }
        [
            {
                label: 'Details',
                fields: ['slug', 'bHasConsumerWebsite', 'address2', 'suburb', 'state', 'postcode']
            },
            {
                label: 'Services',
                fields: ['specialistServices']
            },
            {
                label: 'Test',
                body: function (req, res, record, model, callback) {
                    return callback(null, '<p>Custom render under the test tab.</p>');
                }
            },
            {
                label: 'Multiple fieldsets example',
                body: [
                    {
                        label: 'Discount',
                        fields: ['field1', 'field2', 'field3']
                    },
                    {
                        label: 'Coupon',
                        fields: ['field3', 'field4', 'field5']
                    },
                    {
                        label: 'blah',
                        body: function (req, res, record, model, callback) {
                            return callback(null, '<p>Blah content here.</p>');
                        }
                    }
                ]
            }
        ],
        {
            label: 'Another fieldset below',
            fields: ['something', 'somethingelse']
        }
    ]
}
```

sandytrinh:
@patelswapnil Seeing that we are changing the `overview` DSL I think it might make more sense to render the fields and fieldset as per origin 'Summary' layout. Since it's more compact. See image below.

image

smebberson:
@sandytrinh, I like that complete example of the overview DSL. Looks good and should work well.

lib/formtools/plugins/plugins-helpers.js
```diff
@@ -23,8 +23,8 @@ var defaults = function (options, isEmbeddedDocument) {
...
-    options.form.dateCreated = { label: 'Date created', visible: false };
-    options.form.dateModified = { label: 'Date modified', visible: false };
+    options.form.dateCreated = { label: 'Date created', visible: false, fieldset: 'Summary' };
+    options.form.dateModified = { label: 'Date modified', visible: false, fieldset: 'Summary' };
```
sandytrinh: You can remove the default fieldset here now that we don't need it.

routes/configOverview.js
```diff
@@ -2,61 +2,75 @@
...
+    };
 
-	async.series([
+    async.series([
 
         function (cb) {
 
-            req.linz.config.linz.formtools.overview.summary.renderer(req.linz.record, req.linz.config, function (err, content) {
+            var details = req.linz.config.linz.formtools.overview.details;
+
```
sandytrinh: Do you need to refactor this to work with the new overview DSL?

routes/recordOverview.js
```diff
@@ -3,51 +3,113 @@ var linz = require('linz'),
...
 /* GET /admin/:model/:id/overview */
 var route = function (req, res, next) {
 
-    var locals = {
-        model: req.linz.model,
-        record: clone(req.linz.record.toObject({ virtuals: true})),
-        permissions: req.linz.model.linz.formtools.permissions,
-        formtools: req.linz.model.linz.formtools
-    };
+    function transformDslToOverview (dsl, cb) {
```
sandytrinh: Should move the function `transformDslToOverview` to `L6` so it's a local function outside of `route` to keep the route function smaller. This provides better visibility to the code.

routes/recordOverview.js
```diff
@@ -3,51 +3,113 @@ var linz = require('linz'),
...
-    async.series([
+        if (!(Array.isArray(dsl) && dsl.length)) {
+            return [];
+        }
 
-        function (cb) {
+        var overviewBody = [];
 
-            req.linz.model.linz.formtools.overview.summary.renderer(req.linz.record, req.linz.model, function (err, content) {
+        async.eachSeries(dsl, function (elm, callback) {
```
sandytrinh: what's the variable `elm` stands for?

patelswapnil: we are passing the array (overview.body) to this function. We are looping through each element in the array. Overview.body should be an array which contains objects (fieldsets) and array (tabs)

routes/recordOverview.js
```diff
@@ -3,51 +3,113 @@ var linz = require('linz'),
...
-    var locals = {
-        model: req.linz.model,
-        record: clone(req.linz.record.toObject({ virtuals: true})),
-        permissions: req.linz.model.linz.formtools.permissions,
-        formtools: req.linz.model.linz.formtools
-    };
+    function transformDslToOverview (dsl, cb) {
 
-    async.series([
+        if (!(Array.isArray(dsl) && dsl.length)) {
```
sandytrinh: You haven't cater for when `overview.body` is a function that return html content.

lib/formtools/plugins/plugins-helpers.js
patelswapnil: done.

routes/configOverview.js
patelswapnil: done.

routes/recordOverview.js
patelswapnil: done.

routes/recordOverview.js
patelswapnil: done. renamed `elm` TO `field` to be consistent with rest of code in linz and to be more readable.

routes/recordOverview.js
patelswapnil: done.

patelswapnil:
@sandytrinh , @smebberson this PR is ready for the follow u preview.

lib/formtools/overview.js
```diff
@@ -0,0 +1,145 @@
...
+    // if overview is function execute it and return the content
+    if (typeof overview === 'function') {
+
+        return overview(req, res, record, model, function (err, content) {
+
+            if (err) {
+                return cb(err);
+            }
+
+            return cb(null, {body: content});
```
smebberson: Can you please add the normal spacing here `{ body: content }`.

lib/formtools/overview.js
```diff
@@ -0,0 +1,145 @@
...
+                }
+
+                if (Array.isArray(data) && data.length) {
+
+                    var fieldData = (Array.isArray(field)) ? data : { label: field.label, body: data };
+
+                    overviewBody.push(fieldData);
+                }
+
+                return callback();
```
smebberson: There should be a new line after this one.

lib/formtools/overview.js
```diff
@@ -0,0 +1,145 @@
...
+
+                if (err) {
+                    return cb(err);
+                }
+
+                if (Array.isArray(data) && data.length) {
+
+                    var fieldData = (Array.isArray(field)) ? data : { label: field.label, body: data };
+
+                    overviewBody.push(fieldData);
```
smebberson: There should be a new line after this one.

lib/formtools/overview.js
patelswapnil: done.

lib/formtools/overview.js
patelswapnil: done.

lib/formtools/overview.js
patelswapnil: done.

lib/formtools/overview.js
```diff
@@ -0,0 +1,145 @@
...
+                });
+
+                return callback();
+
+            });
+
+        }
+
+        // make recursive call, if field or field.body is an Array
+        if (Array.isArray(field) || Array.isArray(field.body)) {
```
smebberson: Without being too verbose, can you add a little bit about the DSL you're working with/converting to here? I think this code is a little abstract and could do with a bit more information about what you're working over.

lib/formtools/overview.js
```diff
@@ -0,0 +1,145 @@
...
+
+                    overviewBody.push(fieldData);
+                }
+
+                return callback();
+            });
+
+        }
+
+        if (Array.isArray(field.fields) && field.fields.length) {
```
smebberson: Without being too verbose, can you add a little bit about the DSL you're working with/converting to here? I think this code is a little abstract and could do with a bit more information about what you're working over.

lib/formtools/overview.js
```diff
@@ -0,0 +1,145 @@
...
+        // return if field does not meet any of the above conditions
+        return callback();
+
+    }, function (err) {
+        return cb(err, overviewBody);
+    });
+
+}
+
+function renderFields(record, model, labels, form, fieldsetCaption, fields, callback) {
```
smebberson: Can you add some comments as what this function does? Please use the code blockr Atom editor to provide the information.

lib/formtools/overview.js
```diff
@@ -0,0 +1,145 @@
...
+            if (err) {
+                return cb(err);
+            }
+
+            fieldset.fields.push({
+                label: fieldLabel,
+                value: value
+            });
+
+            return cb();
```
smebberson: New line after this please.

lib/formtools/overview.js
```diff
@@ -0,0 +1,145 @@
...
+    async.eachSeries(fields, function (field, cb) {
+
+        var formField = form[field];
+
+        if (!formField || !formField.visible) {
+            return cb();
+        }
+
+        var renderer = (cellRenderers[formField.type]) ? formField.type : 'text',
+            fieldLabel = labels[field] ? labels[field] : '';
```
smebberson: `fieldLabel` could be empty? Standard practice is to default to the field name itself....

lib/formtools/renderers-cell.js
```diff
@@ -8,7 +8,7 @@ var dateRenderer = module.exports.date = function dateRenderer (date, record, fi
     return callback(null, moment(date).format(linz.get('date format')));
 };
 
-var dateTimeRenderer = module.exports.dateTime = function dateTimeRenderer (date, record, fieldName, model, callback) {
```
smebberson: I think we've discussed this, but why are we moving away from camel casing here?

lib/formtools/templates/index.js
```diff
@@ -1,12 +1,6 @@
 var fs = require('fs'),
     handlebars = require('handlebars');
 
-var documents = fs.readFileSync(__dirname + '/documents.html', 'utf-8'),
-    overview = {};
-
-overview.summary = fs.readFileSync(__dirname + '/overview/summary.html', 'utf-8');
+var documents = fs.readFileSync(__dirname + '/documents.html', 'utf-8');
```
smebberson: I know the previous code had it, but can we move to `path.resolve` for traversing directories, i.e. (`path.resolve(__dirname, 'documents.html')`).

lib/formtools/overview.js
```diff
@@ -0,0 +1,145 @@
+
+var async = require('async'),
+    cellRenderers = require('./renderers-cell');
+
+function body(req, res, record, model, overview, cb) {
```
sandytrinh: @patelswapnil You don't need to pass in `overview` here, as you should have access to it via model, see code below:

```
var overview = model.linz.formtools.overview;
```

routes/configOverview.js
```diff
@@ -1,62 +1,26 @@
...
+        permissions: req.linz.config.linz.formtools.permissions,
+        formtools: req.linz.config.linz.formtools
+    };
 
-	async.series([
+    // Transform overview.body DSL into data object that can be rendered by view
+    linz.formtools.overview.body(req, res, req.linz.record, req.linz.config, req.linz.config.linz.formtools.overview.body, function (err, overviewData) {
 
-        function (cb) {
+        if (!err) {
```
smebberson: I know this probably existed in the old code, but we should be handling the error here. Can you surface it via `res`?

test/formtools.js
```diff
@@ -560,23 +560,6 @@ describe('formtools', function () {
                             label: 'Custom action'
                         }
                     ],
-                    summary: {
```
smebberson: Did you remove tests but not add new ones?

views/mixins/overviewBody.jade
```diff
@@ -0,0 +1,46 @@
+mixin renderOverviewFields(obj, hasTitle, hasSubtitle)
+	section.group
+		div.panel
+			if hasTitle
+				div.panel-heading=obj.label
+			div.panel-body
+				if hasSubtitle
+					h4=obj.label
+				-	if (obj.body)
```
smebberson: Do you need the `-` here? And should the next line be indented?

views/mixins/overviewBody.jade
```diff
@@ -0,0 +1,46 @@
...
+									strong=field.label + ':'
+								span='	' + field.value
+
+
+mixin renderTabs(tabs)
+	div.tabs
+		div
+			ul.nav.nav-tabs.nav-tabs-overview
+				each tab, index in tabs
+					- var tabLink = (tab.label || '').replace(/\s+/g, '-').toLowerCase() + index;
```
smebberson: This stuff shouldn't be within the view itself. Please surface all this code to a route that prepares everything properly for the view.

lib/formtools/overview.js
patelswapnil: done. Refactored overview.js. Added function decription.

lib/formtools/overview.js
patelswapnil: done. Refactored overview.js. Added function decription.

lib/formtools/overview.js
patelswapnil: done. Refactored overview.js. Added function decription.

lib/formtools/overview.js
patelswapnil: done.

lib/formtools/overview.js
patelswapnil: done.

lib/formtools/overview.js
patelswapnil: done.

views/mixins/overviewBody.jade
patelswapnil: no. removed unneccesary `-` charachter. Have refactored the overview jade template.

views/mixins/overviewBody.jade
patelswapnil: done.

lib/formtools/templates/index.js
patelswapnil: done.

lib/formtools/renderers-cell.js
patelswapnil: we renamed few of these variable names in renderers-cell.js to match them with field type returned by `model.linz.formtools.form[field]`.

routes/configOverview.js
patelswapnil: done.

test/formtools.js
patelswapnil: test is out dated. last it was updated 2 years ago. will update it later.

patelswapnil:
@sandytrinh , @smebberson , this PR is ready for follow up review.

lib/formtools/overview.js
```diff
@@ -0,0 +1,183 @@
+
+var async = require('async'),
+    cellRenderers = require('./renderers-cell');
+
+/**
+ * Set unique tabId to each tab in overview.body
+ * @param {Array} overviewBody overview body
```
sandytrinh: You can name `overviewBody` to `body` instead as it is in context.

lib/formtools/overview.js
```diff
@@ -0,0 +1,183 @@
...
+ */
+function setTabId (overviewBody) {
+
+    overviewBody.forEach(function (section, sectionIndex) {
+
+        if (Array.isArray(section)) {
```
sandytrinh: Perhaps you can return early here when `section` is not an array.

lib/formtools/overview.js
```diff
@@ -0,0 +1,183 @@
...
+                tab.tabId = (tab.label).replace(/\s+/g, '-').toLowerCase() + '-' + sectionIndex + '-' + tabIndex;
+
+            });
+        }
+    });
+
+}
+
+/*
+ * Replace each field name in section of overview.body with object containing field label and filed value
```
sandytrinh: Perhaps update this description to `Update section.fields to an array of objects with label and value keys. Label should contain the field label and value should contain the field value.`

lib/formtools/overview.js
```diff
@@ -0,0 +1,183 @@
...
+ * @return {[object]}          Return object: { label: '', fields: [ { label: '', value: '' }, ... ] } to callback
+ */
+function renderFields(record, model, section, callback) {
+
+    var fieldset = {
+        label: section.label,
+        fields: []
+    };
+
+    async.eachSeries(section.fields, function (field, cb) {
```
sandytrinh: Perhaps use `async.mapSeries` here which returns an array you don't need to manually push it into `fields`

lib/formtools/overview.js
```diff
@@ -0,0 +1,183 @@
...
+
+        // recursively goes over nested section which can have repeated object structures
+        return getSections(req, res, record, model, obj, callback);
+
+    }, cb);
+
+}
+
+/*
+ * Get the overview body for record.
```
sandytrinh: Can you please add the word `record` to `a record`?

lib/formtools/overview.js
```diff
@@ -0,0 +1,183 @@
...
+ * @param  {[Object]}   res    Response onject
+ * @param  {[Object]}   record Overview record
+ * @param  {[Object]}   model  Model object
+ * @param  {Object or Array}   section from overvie DSL which can be an Array or an Object
+ * @param  {Function} cb       Callback function
+ * @return {Array}             returns an Array of objects
+ */
+function getSections (req, res, record, model, section, cb) {
+
+    if (typeof section.body === 'function') {
```
smebberson: So, if `section.body` is a `function`, does that mean the resulting code doesn't go through the manipulations below?

lib/formtools/overview.js
```diff
@@ -0,0 +1,183 @@
...
+
+        return section.body(req, res, record, model, function(err, html) {
```
smebberson: We usually define functions with a space between the word `function` and the open bracket `(` (i.e. `function (err, html) {`).

lib/formtools/overview.js
patelswapnil: if `section.body` is a `function`, it will execute that `function` and return the result.

have I answered your question?

smebberson: If `section` is indeed an array, won't this code error?

patelswapnil: No. the following code takes care of it, if `section` is an `array`:

```
async.mapSeries(section, function (obj, callback) {

        // each object (each tabs) in array section of overview body DSL must have label
        if (!obj.label) {
            return cb(new Error('Label is required for each section in overview body.'));
        }

        // recursively goes over nested section which can have repeated object structures
        return getSections(req, res, record, model, obj, callback);

    }, cb);
```

patelswapnil: it will only get to this far if `section` is an array.

smebberson: Not really, my point is, if the user provides an `object` we're allowing the user to provide an incomplete DSL that we will take and fill in the blanks. If however they provide a `function`, that doesn't happen. Shouldn't we be taking the result from that `function` and then running it through the same code to fill in the blanks? I think that is how the other DSL implementations work (i.e. the `grid` object).

lib/formtools/overview.js
patelswapnil: done.

lib/formtools/overview.js
patelswapnil: done. renames `overviewBody` to `sections`.

lib/formtools/overview.js
patelswapnil: done.

lib/formtools/overview.js
patelswapnil: done.

lib/formtools/overview.js
patelswapnil: done. replaces `async.eachSeries` with `async.mapSeries`.

lib/formtools/overview.js
patelswapnil: `if (typeof section.body === 'function')`, will not throw error if `section` is an array.  So all good.

lib/formtools/overview.js
patelswapnil: done.

patelswapnil:
@sandytrinh this PR is ready for the follow up review.

lib/formtools/overview.js
```diff
@@ -0,0 +1,183 @@
...
+
+    sections.forEach(function (section, sectionIndex) {
+
+        if (!Array.isArray(section)) {
+            return;
+        }
+
+        section.forEach(function (tab, tabIndex) {
+
+            // Replace white space from tab.lable with "-", make it lowercase and add the index number at the end.
```
sandytrinh: Typo in `tab.label`

lib/formtools/overview.js
```diff
@@ -0,0 +1,183 @@
...
+            tab.tabId = (tab.label).replace(/\s+/g, '-').toLowerCase() + '-' + sectionIndex + '-' + tabIndex;
+
+        });
+
+    });
+
+}
+
+/*
+ * Update section.fields to an array of objects with label and value keys. Label should contain the field label and value should contain the field value.
```
sandytrinh: Can you please update this sentence to the below for more clarity:
`Update section.fields to an array of objects with 'label' and 'value' keys. 'label' key should contain the field label and 'value' key should contain the field value.`

lib/formtools/overview.js
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * Recursively loops through provided section to:
+ * (1) Render body if section.body is a function (e.g. { label: '', body: '' })
+ * (2) Populate field labels, if section.fields is defined (by converting section.fields into an array of objects. e.g. { label: '', fields: [ { label: '', value: '' }, ... ] })
+ * (3) If section.body is an array, recursively repeats step (1) and (2) (e.g. { label: '', body: [] })
+ * @param  {[Object]}   req    Request object
+ * @param  {[Object]}   res    Response onject
+ * @param  {[Object]}   record Overview record
+ * @param  {[Object]}   model  Model object
+ * @param  {Object or Array}   section from overvie DSL which can be an Array or an Object
```
sandytrinh: Update comment to below:
`section from overview body array which can be an Array or an Object`

lib/formtools/renderers-cell.js
```diff
@@ -156,7 +156,17 @@ var documentArrayRenderer = module.exports.documentArray = function documentArra
 
     });
 
-}
+};
+
+var textRenderer = module.exports.text = function textRenderer(val, record, fieldName, model, callback) {
+
+    if (typeof val === 'undefined') {
```
sandytrinh: You can remove this `if` statement, as `L167` would pass `undefined` as the `val` anyways.

lib/formtools/templates/index.js
```diff
@@ -1,12 +1,6 @@
-var fs = require('fs'),
-    handlebars = require('handlebars');
+var handlebars = require('handlebars'),
+    path = require('path');
 
-var documents = fs.readFileSync(__dirname + '/documents.html', 'utf-8'),
-    overview = {};
-
-overview.summary = fs.readFileSync(__dirname + '/overview/summary.html', 'utf-8');
+var documents = path.resolve(__dirname, '/documents.html');
```
sandytrinh: What's the reason here to replace `fs` with `path` instead? `path` simply provides the path and not the actual content.

routes/configOverview.js
```diff
@@ -1,62 +1,24 @@
...
-                }
-
-				locals.overviewBody = content;
-
-				return cb(null);
-
-			});
+    var locals = {
+        config: req.linz.config,
+        record: req.linz.record,
```
sandytrinh: In `routes/recordOverview.js` `L12` it does a clone of the record, see code below:

```clone(req.linz.record.toObject({ virtuals: true}))```

Perhaps config should follow suit, in record do get manipulated downstream.

File: routes/configOverview.js [hunk 71]
```diff
@@ -1,62 +1,24 @@
...
 
-	], function (err, results) {
+        // Set tabId to each tab in locals.overview.body
+        linz.formtools.overview.setTabId(locals.overview.body);
 
-		res.render(linz.api.views.viewPath('configOverview.jade'), locals);
+    }
 
-	});
+    res.render(linz.api.views.viewPath('configOverview.jade'), locals);
```
sandytrinh: Can you please add `return` at the beginning of this line of code to follow our standard practice?

routes/recordOverview.js
```diff
@@ -3,75 +3,26 @@ var linz = require('linz'),
...
-        function (cb) {
-
-            if (!req.linz.model.linz.formtools.overview.body) {
-
-                return cb(null);
-
-            }
-
-            req.linz.model.linz.formtools.overview.body(req, res, req.linz.record, req.linz.model, function (err, content) {
+            model: req.linz.model,
```
sandytrinh: Can you please remove the extra indentation here?

test/formtools.js
sandytrinh: @smebberson The unit tests was created prior to vagrant setup. We need an environment to run the tests. Also the tests are over 2 years old so there are a lot of breaking changes included since, so we'll need to set aside time to get this up and running again.

lib/formtools/templates/index.js
patelswapnil: done. reverted to original code then replaced `__dirname + '/overview/summary.html'` with `path.resolve(__dirname, 'documents.html')`.

lib/formtools/overview.js
patelswapnil: done.

lib/formtools/overview.js
patelswapnil: done.

lib/formtools/overview.js
patelswapnil: done.

lib/formtools/renderers-cell.js
patelswapnil: done.

routes/configOverview.js
patelswapnil: done. used `clone(req.linz.record)`as config record does not have `toObject` function.

routes/configOverview.js
patelswapnil: done.

routes/recordOverview.js
patelswapnil: done.

patelswapnil:
@sandytrinh this PR has ready for follow up review.

PR #104, opened 2016-08-23
Various bug fixes

sandytrinh:
This PR fixes the following:
- [x] Fix edit page, where it's compulsory to add `embeddedDocument` plugin to any embedded schema for a document. This is no longer a requirement and standard embedded document can be use that doesn't require to be render in the edit page.
- [x] Fix namespace for model form configurations in recordChange middleware.
- [x] Update `linz.api.formtools.grid` function so grid data is retrieved based on a user
- [x] Add `linz.utils.arrayEquals`
- [x] Add a transpose function (similar to transform but works server to client, rather than client to server).
- [x] Fix Reset Password
 - [x] Add `middleware.linz` to reset password route so `req.linz` namespace is available
 - [x] Removed login button from 'Forgotten password' confirmation screen as it's not required
 - [x] Clear `res.clearCookie('linz-return-url') ` so it doesn't redirect back to 'Reset Password' page after successful login
- [x] Refactore data model compare function
 - [x] Should only flag fields that are changed
 - [x] Update text highlighting for compare view. It should only highlight text change for 'Text' fields. This should remove the confusion on what's added and what's removed.
 - [x] Fix incorrect date where it also shows the current date for both records
 - [x] Render date in local date time
 - [x] Update some of the function to ES6 arrow functions
 - [x] Add group heading "Compare with previous versions" for the dropdown list under "Compare with latest" 

__Previous__
image

__Updated__
image

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -18,5 +18,7 @@
         "no-trailing-spaces": 1,
         "no-process-env": 0,
         "no-process-exit": 0
-    }
+    },
+    # if you are receiving errors for eslint, comment out the line below or upgrade to 7.2.2+
+    "extends": "eslint:recommended"
 }
--- a/lib/api/formtools/grid.js
+++ b/lib/api/formtools/grid.js
@@ -17,33 +17,38 @@ function renderToolbarItems (req, res, modelName, cb) {
         throw new Error('modelName is required and must be of type String and cannot be empty.');
     }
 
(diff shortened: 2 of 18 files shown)


sandytrinh:
@smebberson This is ready for review.

smebberson:
@sandytrinh, this looks good! Please merge.

PR #105, opened 2016-09-06
Overview action model fix

sandytrinh:
This is a quick PR that fixes the overview model action where `modal.active` is always set to true if modal is defined in `overview.actions`.

Now overview will respect `modal.active` value and default it to `true` if it is `undefined`.


sandytrinh:

sandytrinh:
This has been review by Scott.

PR #108, opened 2016-12-13
Few Bug fixes

patelswapnil:
## This PR will fix following bugs:

 - [x] Number value `0` and `1` are rendered as `No` and `Yes` in the gird (List page)
 - [x] Pagination does not work when model does not include filter

### Before bug fix:
 -
 screen shot 2016-12-14 at 10 10 14 am

 - Pagination does not update the list of records

### After bug fix: 
 -
 screen shot 2016-12-13 at 4 20 17 pm

 - Now, Pagination works as it should on model that does not include filter.

### DIFF
--- a/lib/formtools/renderers-cell.js
+++ b/lib/formtools/renderers-cell.js
@@ -204,6 +204,10 @@ var defaultRenderer = module.exports.default = function defaultRenderer(val, rec
 
     }
 
+    if (typeof val === 'number') {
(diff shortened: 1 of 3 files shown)


patelswapnil:
@smebberson, this PR is ready for review.

smebberson:
@patelswapnil, looks good. Please merge and make a new tag :)

PR #110, opened 2017-01-09
Remove old options

allanchau:


allanchau:
@smebberson argh how did that get in there? Removed now

lib/formtools/plugins/plugins-helpers.js
```diff
@@ -29,10 +29,6 @@ var defaults = function (options, isEmbeddedDocument) {
     // set up fields to an empty object as default
     options.fields = options.fields || {};
```
smebberson: I think this could go too? We're not using it for anything else as far as I can see or search for... Please try removing it and running the code to make sure it doesn't break.

allanchau: Yeah all good

allanchau:
@smebberson options.fields has been removed now

PR #116, opened 2017-03-10
List renderer

allanchau:
This PR integrates adds an option to render a `list` view rather than the default `grid` view.

## Setup

- [ ] Make sure your project has been updated with the `grid` -> `list` and `columns` -> `fields`.

## Testing

- [x] With an existing project, open up a list in your browser. Nothing should change.
- [x] Add a new renderer property to `list`, i.e. `list.renderer` with the value `linz.formtools.listRenderers.default`. Nothing should change.
- [x] Add a new renderer property to `list`, i.e. `list.renderer` with the value `linz.formtools.listRenderers.grid`. Nothing should change.
- [x] Add a new renderer property to `list`, i.e. `list.renderer` with the value `linz.formtools.listRenderers.list`. You should now see a list view.


smebberson:
@allanchau, if you take a look at widgets, you'll note that they export a function which accepts `attrs` and returns a `function` that you pass to Linz.

I think that perhaps we should do the same for the new `list` renderers. The reason being, it would allow us to customise things such, in the case of the `list` list renderer, the properties that are rendered to the definition list.

What do you think?

allanchau:
@smebberson Would this work like this?
```javascript
/**
 * List list renderer.
 * @param  {Object} attrs Express response object.
 * @return {Function}     Returns a function to render the template.
 */
function list (attrs) {

    const defaults = {};

    utils.merge(defaults, (attrs || {}));

    return (res) => {

        return res.render(linz.api.views.viewPath('list.jade'), defaults);

    }

}

// Calling the function.
list(data)(req);
```

I can see it being useful for things like ensuring there is always a title, but isn't that considered a side effect?

smebberson:
@allanchau, it should be more like this:

```
/**
 * List list renderer.
 * @param  {Object} attrs Express response object.
 * @return {Function}     Returns a function to render the template.
 */
function list (attrs) {

    return (res, data) => {

        return res.render(linz.api.views.viewPath('list.jade'), utils.merge(data, attrs || {}));

    }

}
```

Then, when defining a renderer in the list DSL you'd execute the first `function`, but not the `return` `function`, like so:

```
{
    renderer: list.formtools.listRenderers.list()
}
```

We should program the `list` renderer to support an array of field names that can be used to filter the definition list generated for each record, i.e:
```
{
    renderer: list.formtools.listRenderers.list({ fields: ['name', 'date', 'logo', 'whatever'] })
}
```

smebberson:
@allanchau, I'm an idiot. It's already being filtered by `list.fields`. It should only include those fields in the definition list. Ignore those last comments please :(

smebberson:
@allanchau, looking good. Could you please make the following CSS related changes:

- Set all `dt` tags to have a `font-weight` of `600`.
- Set `entry-title.h1` to have a `font-size` of `28px`.

With that, we'll be done here :)

smebberson:
@allanchau, if we're changing the CSS, we should automatically cache-bust the file. Please add `?v1` to the place (somewhere in a jade file) which makes reference to the CSS file. This will ensure that the latest CSS is automatically picked up.

So, you'll need to add `?v1` to https://github.com/linzjs/linz/blob/8fb44e19252c3f9ea48b4cd424bb80ef140b5116/views/layout.jade#L19 and the same in `layout-preauth.jade`...

Make sense?

smebberson:
@allanchau. Nice. Merging!

PR #117, opened 2017-04-07
Modal record action

allanchau:
This PR adds a modal option to record-actions that is currently available with group-actions.

## Setup

- You will need an existing record action.
- You will need the associated views in your project.

## Related PRs

- None.

## Testing

- [x] With an existing record action, add the `modal: true` property.
- [x] Trigger the action and check that the modal correctly renders (This is dependant on the view so if anything pops up it should be working).

### DIFF
--- a/docs/models_list.rst
+++ b/docs/models_list.rst
@@ -30,7 +30,7 @@ list.actions
 
 - ``label`` is the name of the action.
 - ``action`` is the last portion of a URL, which is used to perform the action.
-- ``useModal`` optionally render the results in a modal view.
+- ``modal`` optionally render the results in a modal view.
 
 For example::
 
@@ -38,7 +38,7 @@ For example::
     {
       label: 'Import people',
       action: 'import-from-csv',
-      useModal: true
+      modal: true
     }
   ]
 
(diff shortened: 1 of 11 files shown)


allanchau:
@smebberson Ready for a review :) Docs and version number have also been updated.

smebberson:
@allanchau, could you please resolve the conflicts.

docs/models_list.rst
```diff
@@ -30,13 +30,15 @@ list.actions
 
 - ``label`` is the name of the action.
 - ``action`` is the last portion of a URL, which is used to perform the action.
+- ``useModal`` optionally render the results in a modal view.
```
smebberson: Can we change this to just `modal`, which is more inline with other current DSLs in Linz (such as overview actions and list actions).

allanchau:
@smebberson Modals have been fixed 🥇

public/js/model/index.js
```diff
@@ -166,6 +166,18 @@
        return false;
     });
 
+    // bind record action buttons
+    $('[data-linz-control="record-action"]').click(function () {
+
+        var queryObj = $(this),
+            url = queryObj.attr('href');
+
+        $('#recordActionModal').modal().find('.modal-dialog').load(url, function () {});
```
smebberson: @allanchau, Does this need to be updated to read `.find('.modal-content')`?

PR #119, opened 2017-04-11
Document fix

allanchau:
Prevents an indexing bug when trying to wrap `list` in a `(user, callback)` function.

### Setup

  - [ ] Add something like this to your list model:
```
module.exports = (user, callback) => {

    return callback(null, {
        fields: ...,
        filters: ...,
        export: ...,
        sortBy: ...,
        renderer: linz.formtools.listRenderers.list
    });

};
```

### Testing

  - [x] There should be no visible changes.

### Notes

  - As a bonus, I have also added some basic Github templates :).

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,11 @@
+### Expected behaviour
+
+
+
+### Actual behaviour
+
+
+
+### Steps to reproduce the behaviour
+
+  1.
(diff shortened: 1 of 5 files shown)


smebberson:
@allanchau, could you please update this PR to not state the version. It should just be `## Unreleased` for the CHANGELOG (unless there already is one).

allanchau:
@smebberson All done :)

smebberson:
@allanchau, oh, you have to remove the version bump too :)

PR #120, opened 2017-04-12
Allow html for custom list renderers

allanchau:
This PR fixes the previous list renderers and enables renderer customisation.

### Testing

- [x] Nothing should change.
- [x] Add the following custom renderer (or provide your own):
  ```
        renderer: (data, callback) => {

            const records = data.records;
            let html = '';

            for (const record in records) {
                html += `<p>${record}</p>`;
            }

            return callback(null, html);

        }
  ```
  You should see an updated list.


allanchau:
@smebberson Fixed!

smebberson:
@allanchau, could you please update the changelog with bug fix information? Nothing breaking here is there?

allanchau:
@smebberson Not that I can think of, it was broken before so you wouldn't of been able to provide a custom renderer.

smebberson:
@allanchau, okay, that's working better. Just thinking this through, there is quite a lot of logic in the actions part of these renderers.

If a developer were to implement this and want to use a new renderer, I think we should keep that logic within Linz. What do you think?

Can we create a new API hook that Linz uses itself to render the actions HTML for a particular record?

allanchau:
@smebberson Yeah sounds good, would we have to create new custom renderers for the record actions? That way it can be customised if needed.

Also, would it be worth splitting this functionality into a different PR? These commits are more of a bug fix, whereas adding new api sounds like a feature?

allanchau:
@smebberson I've started a branch in my own fork, but what do you think about https://github.com/linzjs/linz/issues/121 ?

smebberson:
@allanchau, all the same can be fine. We can do a minor release, without any issues, and include bug fixes.

smebberson:
@allanchau, I really like #121, can we do that one next however, and just get through this one?

smebberson:
By custom renderers for the record actions, do you mean, just hiving off the record actions HTML into it's own seperate view (i.e. `partials/action-record.hbs`) from #121? If so, then yes, that'd be good.

allanchau:
yeah exactly.

smebberson:
@allanchau, we still need an API that will render that for us, and return the HTML, just like we did with the list renderer...

routes/modelIndex.js
```diff
@@ -79,6 +78,41 @@ var route = function (req, res, next) {
             plural: req.linz.model.linz.formtools.model.plural
         },
         modelQuery: JSON.stringify(req.linz.model.formData)
+    };
+
+    const renderRecordAction = data.model.list.recordActions.renderer || recordActionRenderers.defaultRenderer;
+
+    data.records.forEach((record, index) => {
+
+        renderRecordAction({
```
allanchau: @smebberson I think we need something better here, calling `app.render` per record will cause a performance hit?

smebberson: @allanchau, will it? Why don't you time it before and after the change... Just remember, you might need to run with `NODE_ENV=production` or something like that. Many Node.js renderers operate different based on that ENV variable, for example, some don't use template caching while in DEBUG mode.

allanchau: @smebberson Yeah that seems a bit quicker, I think it may also have been the networking in virtualbox. The other screens seem slow sometimes too.

views/modelIndex/grid.jade
```diff
@@ -0,0 +1,25 @@
...
+                        for field in model.list.fields
+                            th= field.label
+                tbody
+                    for record, index in records
+                        tr
+                            td= from + index + 1
+                            if model.list.groupActions && model.list.groupActions.length
+                                td
+                                    input(type="checkbox", name="ids[]", value="#{record._id}", data-linz-control="checked-record")
+                            if permissions.canDelete !== false || permissions.canEdit !== false || model.list.recordActions.length
```
smebberson: This logic doesn't seem right. I think it should just be `if model.list.recordActions.length`. For a couple of reasons, there might be some custom `recordActions` but they'll never be rendered if `permissions.canDelete === false`.

Also, we're already dealing with `permissions.canDelete` and `permissions.canEdit` within `action-record.jade`...

allanchau: @smebberson Actually now that I look at it, we can remove the check altogether.
We have moved the record-actions into a separate template so if any logic is needed, it can be managed there?

smebberson: Well, in fact, we need the basic `if` statement. We'll need to change L10 of this file too. If there is no `recordActions` we don't want the `th` and `td`. So we'll need the check for that...

allanchau:
@smebberson ok, done now :)

PR #123, opened 2017-04-28
Fix modal record actions

allanchau:
It seems `action.modal` is always present. `action.modal.active` is the boolean that controls whether a user has set `modal: true`


allanchau:
@smebberson This should fix all record actions appearing as modals.

PR #126, opened 2017-05-02
Render

allanchau:
This PR adds the ability to render a view in the context of Linz.

### Testing

- [ ] Pass some data to the render function and see if it works.
```javascript
const options = { page: homepage };

linz.api.views.render(options, (err, html) => {

    if (err) {
        return res.send(err);
    }

    return res.send(html);

});
```
- [ ] By default, nothing should change.
- [ ] Replace the callback function with `res`, it should still work.
- [ ] Add a new key to the options `template: 'wrapper-preauth'`. The view should now display without the navigation components.

### DIFF
--- a/lib/api/views.js
+++ b/lib/api/views.js
@@ -1,5 +1,7 @@
-var linz = require('../../'),
-    path = require('path');
+'use strict';
+
+const linz = require('../../');
+const path = require('path');
 
 /**
  * Return the path to Linz's view directory, appending the specific view if passed in.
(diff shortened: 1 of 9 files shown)


lib/api/views.js
```diff
@@ -1,16 +1,45 @@
...
+ * @return {String}            Renders and returns the template HTML.
+ */
+const render = (data, callback) => {
+
+    const appLocals = {
+        linzNavigation: linz.get('navigation'),
+        template: 'wrapper'
+    };
+
+    const locals = utils.merge(appLocals, data || {});
```
smebberson: @allanchau, try this... `const locals = Object.assign(appLocals, clone(data || {}));` (you'll need to require `clone` of course).

PR #127, opened 2017-05-08
Add option to display record actions under the overview

allanchau:
This PR adds the option to display record actions underneath an overview.
screen shot 2017-05-08 at 1 40 07 pm


allanchau:
@smebberson Ready for review

smebberson:
@allanchau, so this is a completely new Overview DSL entry, called `footerActions`? It would seem as though you need to run these through the `actionsDefaults` function, just as we do to `recordActions` and `groupActions`.... https://github.com/linzjs/linz/blob/master/lib/formtools/plugins/plugins-helpers.js#L102

smebberson:
@allanchau, I was originally thinking that all of the `recordActions` would be put down there. I like this option however, that you can supply different options down the bottom.

How would one put the edit/delete standard Linz actions down there too?

allanchau:
@smebberson We could add it to the template. They will only display if the user has the right permissions anyway.

smebberson:
@allanchau, yeah, correct regarding permissions. Let's put them down there. So it will be the default actions that Linz sets up for a record, and then the footer actions. Cool?

smebberson:
@allanchau, cool, let's move forward with that :)

views/recordOverview.jade
```diff
@@ -56,6 +56,15 @@ block content
                 if overview.versions
                     .overview-versions
                         div!= overview.versions
+    if formtools.overview.footerActions && formtools.overview.footerActions.length
+        .container
+            .row
+                .col-xs-12
+                    .linz-overview-footer-actions
+                        .model-actions
+                            for action in formtools.overview.actions
```
smebberson: Do you need a second loop for the `footerActions`?

smebberson:
@allanchau, I just added this but only the `actions` are coming out below...

```
actions: [
            {
                label: 'Record Action One',
                action: 'record-action-one'
            },
            {
                label: 'Record Action Two',
                action: 'record-action-two'
            }
        ],
        footerActions: [
            {
                label: 'Footer Action One',
                action: 'footer-action-one'
            },
            {
                label: 'Footer Action Two',
                action: 'footer-action-two'
            }
        ]
```

views/recordOverview.jade
```diff
@@ -56,6 +56,27 @@ block content
...
+                                        span.edit
+                                            a.btn.btn-primary(href='' + linz.api.url.getAdminLink(model, 'edit', record._id), data-linz-control="edit", data-linz-disabled=((record.edit && record.edit.disabled) ? 'true' : false), data-linz-disabled-message=((record.edit && record.edit.disabled) ? record.edit.message : false))
+                                                i.fa.fa-edit
+                                                | Edit
+                                    if permissions.canDelete !== false
+                                        span.delete
+                                            a.btn.btn-default(href='' + linz.api.url.getAdminLink(model, 'delete', record._id), data-linz-control="delete", data-linz-disabled=((record.delete && record.delete.disabled) ? 'true' : false), data-linz-disabled-message=((record.delete && record.delete.disabled) ? record.delete.message : false))
+                                                i.fa.fa-trash-o
+                                                | Delete
+                            for action in formtools.overview.footerActions
```
smebberson: @allanchau, these need to be programmed to support disabled actions, just like we have in the actions at the top...

routes/recordOverview.js
```diff
@@ -91,6 +91,26 @@ var route = function (req, res, next) {
 
         });
 
+        // Define default overview action modal settings in a format that jade can access easily.
+        req.linz.model.linz.formtools.overview.footerActions.forEach((action) => {
```
smebberson: In fact, this is duplicate code. Can you create a function to do this and run both the `footerActions.forEach` and `actions.forEach` against it?

public/src/css/includes/overview.less
```diff
@@ -158,3 +158,29 @@
   }
 
 }
+
+.linz-overview-footer-actions {
+
+  margin-top: 3rem;
```
smebberson: @allanchau, can you remove this? I think it's too much padding and we don't need it.

lib/api/formtools/actions.js
```diff
@@ -0,0 +1,78 @@
+'use strict';
+
+/**
+ * Parse some action disabled properties.
+ * @param  {Object} record Mongoose model record.
+ * @param  {Array|Object} actions Actions to parse.
+ * @return {Array} Parsed actions.
```
smebberson: This doesn't return an `array`? It returns a `Promise`?

lib/api/formtools/actions.js
```diff
@@ -0,0 +1,78 @@
...
+ */
+const parseDisabledProperties = (record, actions) => new Promise((resolve, reject) => {
+
+    const parsedActions = [].concat(actions);
+    const actionsToBeParsed = [];
+
+    parsedActions.forEach((action) => {
```
smebberson: I feel it would be cleaner and easier to read by using `parsedActions.filter` first to filter based on `action.disabled`. It could remove that `if` statement below.

allanchau:
@smebberson All fixed now.

PR #128, opened 2017-05-31
Add query parameters to the list renderer data

allanchau:
This PR adds query parameters to the list renderer data.


allanchau:
@smebberson Ready for review.

PR #135, opened 2017-07-24
Navbar icons

allanchau:
Add the ability to provide icons to the navbar.


allanchau:
@smebberson Ready for a review.

smebberson:
@allanchau, could you please add a head and some basic `lorem ipsum` for the _Contact us_ and _FAQ_ pages.

smebberson:
@allanchau, when I log in as a `member` level user, not `admin` level user - I can see the _Subscription_ navigation menu but I can't access it. Can we just remove it based on user level?

allanchau:
@smebberson I've updated this PR with additional navigation and body tag attribute configurations.

PR #137, opened 2017-07-28
Scripts styles

allanchau:
This PR adds a `scripts` and `styles` option to Linz that allows custom scripts and tags within each template.


routes/mergeDataConflictGuide.js
```diff
@@ -3,7 +3,10 @@ var linz = require('../');
 /* GET /admin/merge-data-conflict-guide */
 var route = function (req, res) {
 
-    res.render(linz.api.views.viewPath('merge-data-conflict-guide.jade'), {});
+    res.render(linz.api.views.viewPath('merge-data-conflict-guide.jade'), {
+        scripts: res.locals.scripts,
```
smebberson: Doesn't `res.render` automatically get `res.locals` made available to it?

views/partials/foot.jade
```diff
@@ -0,0 +1,4 @@
+if scripts && scripts.length
+	each script in scripts
+		if !script.inHead
+			script(async=script.async crossorigin=script.crossorigin defer=script.defer integrity=script.integrity src=script.src text=script.text type=script.type)
```
smebberson: I presume this doesn't cause issues if some of the values are missing?

routes/mergeDataConflictGuide.js
allanchau: I'm not really sure, the docs don't seem that clear to me http://expressjs.com/en/api.html#res.locals

views/partials/foot.jade
allanchau: It shouldn't, jade should strip out all undefined properties
`<script crossorigin="anonymous" integrity="sha256-k2WSCIexGzOj3Euiig+TlR8gA0EmPjuc79OEeY5L45g=" src="//code.jquery.com/jquery-3.2.1.slim.min.js"></script>`

allanchau:
@smebberson I think it would be better to wait a bit before we deprecate the current defaults.
I'm planning to do a bit of a cleanup of the current Linz scripts and move them into the new scripts middleware, that way we can dynamically load scripts per page and make Linz more performant 💯 .

smebberson:
@allanchau, could you please add some documentation here? It should go into the https://github.com/linzjs/linz/blob/master/docs/defaults.rst file.

allanchau:
@smebberson Ready for another review.

routes/configEdit.js
```diff
@@ -1,22 +1,52 @@
...
+
+        Promise.all([
+            setTemplateScripts(req, res, [
+                {
+                    src: '//cdnjs.cloudflare.com/ajax/libs/handlebars.js/4.0.10/handlebars.min.js',
+                    integrity: 'sha256-0JaDbGZRXlzkFbV8Xi8ZhH/zZ6QQM0Y3dCkYZ7JYq34=',
+                    crossorigin: 'anonymous',
+                },
+                {
+                    src: `${linz.get('admin path')}/public/js/jquery.binddata.js`,
```
smebberson: Can this be loaded from a CDN?

routes/modelExport.js
```diff
@@ -1,9 +1,33 @@
-var linz = require('../');
+'use strict';
+
+const linz = require('../');
+const setTemplateScripts = require('../lib/scripts');
```
smebberson: This process where you load the methods, then do `Promise.all`... could that all be put into the API? Anything that we do in Linz, I want programmers to be able to do also. At the moment, they can't access this functionality simply. If we add it to the API and we use it, they'll be able to do the same. This will affect a lot of the PR.

routes/recordEdit.js
```diff
@@ -44,15 +46,55 @@ var route = function (req, res, next) {
...
+                crossorigin: 'anonymous',
+            },
+            {
+                src: `${linz.get('admin path')}/public/js/jquery.binddata.js`,
+            },
+            {
+                src: `${linz.get('admin path')}/public/js/documentarray.js`,
+            },
+            {
+                src: `${linz.get('admin path')}/public/js/deep-diff-0.2.0.min.js`,
```
smebberson: Is this on a CDN?

routes/recordEdit.js
```diff
@@ -44,15 +46,55 @@ var route = function (req, res, next) {
...
+            },
+            {
+                src: `${linz.get('admin path')}/public/js/json2.js`,
```
smebberson: Is this on a CDN?

views/partials/foot.jade
```diff
@@ -0,0 +1,8 @@
...
+script.
+	CKEDITOR.replaceClass = null; // Disable replacing by class
+	CKEDITOR.disableAutoInline = true;
+script. linz.loadLibraries('#{linz.get('admin path')}');
```
smebberson: This dies on the login page, right? Is there something we can do about that?

views/partials/head.jade
```diff
@@ -0,0 +1,12 @@
...
+	meta(http-equiv='X-UA-Compatible' content='IE=edge')
+	meta(name='viewport' content='width=device-width, initial-scale=1')
+	title= pageTitle
+	if styles && styles.length
+		each style in styles
+			link(as=style.as crossorigin=style.crossorigin href=style.href hreflang=style.hreflang media=style.media rel=style.rel sizes=style.sizes title=style.title type=style.type)
+	if scripts && scripts.length
+		each script in scripts
+			if script.inHead
+				script(async=script.async crossorigin=script.crossorigin defer=script.defer integrity=script.integrity src=script.src text=script.text type=script.type)
```
smebberson: I know this is an edge case, and I'm willing to let it go, but is there an easy way to allow the programmer to pass in other attributes and have them output with the tag? For example, how would a programmer create a `script` tag with the attribute `data-custom-attr="something"`? Can you look into if we can expose this someway in Jade?

lib/scripts.js
```diff
@@ -0,0 +1,86 @@
...
+        {
+            crossorigin: 'anonymous',
+            integrity: 'sha256-1hjUhpc44NwiNg8OwMu2QzJXhD8kcj+sJA3aCQZoUjg=',
+            src: '//cdnjs.cloudflare.com/ajax/libs/moment.js/2.18.1/moment.min.js',
+        },
+        {
+            src: `${linz.get('admin path')}/public/js/utils.js`,
+        },
+        {
+            src: `${linz.get('admin path')}/public/ckeditor/ckeditor.js`,
```
allanchau: @smebberson Yeah, but I'm not sure what customisations we have.

lib/scripts.js
```diff
@@ -0,0 +1,86 @@
...
+            src: '//maxcdn.bootstrapcdn.com/bootstrap/3.3.7/js/bootstrap.min.js',
+
+        },
+        {
+            crossorigin: 'anonymous',
+            integrity: 'sha256-1hjUhpc44NwiNg8OwMu2QzJXhD8kcj+sJA3aCQZoUjg=',
+            src: '//cdnjs.cloudflare.com/ajax/libs/moment.js/2.18.1/moment.min.js',
+        },
+        {
+            src: `${linz.get('admin path')}/public/js/utils.js`,
```
allanchau: @smebberson yeah I tried that, linz isn't defined though?

lib/scripts.js
```diff
@@ -0,0 +1,86 @@
...
+        },
+        {
+            src: `${linz.get('admin path')}/public/ckeditor/ckeditor.js`,
+        },
+        {
+            src: `${linz.get('admin path')}/public/js/modernizr.custom.min.js`,
+        },
+        {
+            src: `${linz.get('admin path')}/public/js/multiselect.min.js`,
```
allanchau: I didn't find a compatible one here: https://cdnjs.com/libraries/bootstrap-multiselect and I think we should require `SRI` that jsdelivr doesn't seem to provide?

lib/scripts.js
```diff
@@ -0,0 +1,86 @@
...
+        },
+        {
+            src: `${linz.get('admin path')}/public/js/noty.min.js`,
```
allanchau: I couldn't find one.

lib/styles.js
```diff
@@ -0,0 +1,68 @@
...
+            href: '//maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css',
+            integrity: 'sha384-wvfXpqpZZVQGK6TAh5PVlGOfQNHSoD2xbE+QkPxCAFlNEevoEH3Sl0sibVcOQVnN',
+            rel: 'stylesheet',
+        },
+        {
+            href: `${linz.get('admin path')}/public/css/multiselect.min.css`,
+            rel: 'stylesheet',
+        },
+        {
+            href: `${linz.get('admin path')}/public/css/noty.css`,
```
allanchau: I couldn't find one.

lib/scripts.js
```diff
@@ -0,0 +1,86 @@
...
+        },
+        {
+            src: `${linz.get('admin path')}/public/js/ui.js`,
+        },
+        {
+            src: `${linz.get('admin path')}/public/js/validator.min.js`,
+        },
+    ];
+
+    if (linz.app.locals.adminJSFile) {
```
allanchau: I think we should leave it deprecated for a few versions?

lib/styles.js
```diff
@@ -0,0 +1,68 @@
...
+            rel: 'stylesheet',
+        },
+        {
+            href: `${linz.get('admin path')}/public/css/linz.css?v2`,
+            rel: 'stylesheet',
+        },
+    ];
+
+    if (linz.app.locals.adminCSSFile) {
```
allanchau: Same as above, I think we should leave it deprecated for a few versions.

routes/configEdit.js
allanchau: Same as above, I couldn't find one on cdnjs

routes/modelExport.js
allanchau: @smebberson How would a programmer use it though?

I actually though about something like this to set `res.locals` and expose that through the api. That is probably best done in another PR though since it would require refactoring of a lot of code that isn't related to `scripts` and `styles`?

views/partials/head.jade
allanchau: @smebberson I don't think Jade allows this, but I'll add it in the v1 project as I know Marko allows it.

views/partials/foot.jade
allanchau: @smebberson Hrmm, from what I can see this just loads the bootstrap-datetimepicker for date fields. I have done some testing and it seems that it isn't working in browsers other than chrome.

I can just remove it and fix up the datepicker in another pr? I'm thinking we just load it with the common scripts and rewrite the datepicker? What do you think?

lib/scripts.js
smebberson: @allanchau, well, download the same version from the CDN, do a quick diff using Kaleidoscope. You shouldn't find any changes though.

lib/scripts.js
smebberson: @allanchau, okay, I don't want to leave it in there forever though....

docs/defaults.rst
```diff
@@ -71,3 +71,64 @@ The function should return an array of objects containing attributes in a name/v
...
+      {
+        crossorigin: 'anonymous',
+        integrity: 'sha256-5YmaxAwMjIpMrVlK84Y/+NjCpKnFYa8bWWBbUHSBGfU=',
+        src: '//cdnjs.cloudflare.com/ajax/libs/bootstrap-datetimepicker/4.17.47/js/bootstrap-datetimepicker.min.js',
+      },
+    ]));
+
+  }
+
+``res.locals.scripts`` contains all the scripts used by Linz, be careful when removing/updating these as it could break functionality within Linz.
```
allanchau: Done ```The script objects can contain an additional ``inHead`` boolean option to optionally load the script in the head tag.```

docs/defaults.rst
```diff
@@ -71,3 +71,64 @@ The function should return an array of objects containing attributes in a name/v
...
+        crossorigin: 'anonymous',
+        href: '//cdnjs.cloudflare.com/ajax/libs/bootstrap-datetimepicker/4.17.47/css/bootstrap-datetimepicker.min.css',
+        integrity: 'sha256-yMjaV542P+q1RnH6XByCPDfUFhmOafWbeLPmqKh11zo=',
+        rel: 'stylesheet',
+      },
+    ]));
+
+  }
+
+``res.locals.styles`` contains all the styles used by Linz, be careful when removing/updating these as it could break functionality within Linz.
```
allanchau: Done

lib/scripts.js
allanchau: Do we use any plugins?
screen shot 2017-08-08 at 3 45 17 pm

routes/modelExport.js
smebberson: My vision for Linz is to become one big API. All of the pages that we deliver should use that API. Down the track, I want to allow someone to override the HTML for any page within Linz. Having the ability to spit out dependent CSS and JS in the same fashion as the rest of Linz works would be important. Does that make sense?

I don't think we need to expose `res.locals` via the API, because programmers have access to Express and can set it themselves?

allanchau: So `linz.api.views.getScripts(req, res)` would get the list of scripts for that route?

Linz updates the `res.locals` after the user has access to it, by exposing it at the last second we can allow the user to edit whatever they want on the page. In v1 we would update all the Marko templates to use `out.global.somelocalproperty` which contains req, res, app, app.locals, and res.locals http://markojs.com/docs/express/#usage

views/partials/foot.jade
smebberson: Let's leave it for now.

views/partials/head.jade
smebberson: Pug has `&attributes` which explodes an object into attribute key pairs... but not in the version we have. Oh well, we do it down the track.

routes/modelExport.js
smebberson: Yeah, something like:

```
linz.api.views.getScripts = function (req, res, scripts = []) {
    return setTemplateScripts(req, res, scripts);
}

linz.api.views.getStyles = function (req, res, styles = []) {
    return setTemplateStyles(req, res, styles);
}
```

Then you'd use it like so in `routes/modelExport.js` for example:
```
Promise.all([linz.api.views.getScripts(req, res, [{ src: '/something.js' }]), linz.api.views.getScripts(req, res, [{ src: '/something.js' }])])
   .then([scripts, styles] => {
      return res.render();
   });
```

Make sense?

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 - Allow custom scripts and styles through the `scripts` and `styles` Linz options. These should be functions that take `req` and `res` objects and return a promise that resolves with an array of scripts and styles. The array should contain objects with the same properties as the HTML equivalents.
 - Now passing `form` and `user` to `transform` function.
 - Now passing `record` to `transpose` function.
+- Add the ability to set an `alwaysOn` property for filters. This should be a boolean and will remove the filter from the dropdown, hide the close button, and automatically show the filter.
```
smebberson: @allanchau, this isn't supposed to be in this PR is it?

allanchau: oh no, how did that get in there?

PR #139, opened 2017-07-31
Add form and user parameters to the transform functions

allanchau:
This PR adds additional parameters to the model `transform` function.


smebberson:
@allanchau, could you please update the documentation. It was a little [light-on](https://linzjs.readthedocs.io/en/latest/models_form.html#field-name-transform) for these functions anyway :(

allanchau:
@smebberson I've added the additional parameters, to the `transpose` function.

PR #143, opened 2017-08-08
#1 - Always on filters

allanchau:
This PR adds the ability to set an `alwaysOn filter`.

### DIFF
--- a/docs/models_list.rst
+++ b/docs/models_list.rst
@@ -126,11 +126,14 @@ list.filters
 
   filters: {
     dateModified: {
-      filter: linz.formtools.filters.dateRange
+      alwaysOn: true,
+      filter: linz.formtools.filters.dateRange,
     }
   }
 
-The following will allow your model to be easily filtered by a date range filter, on the ``dateModified`` property. For a complete list of the filters available see https://github.com/linzjs/linz/tree/master/lib/formtools/filters.
+The following will allow your model to be easily filtered by a date range filter, on the ``dateModified`` property. You can also set `alwaysOn: true` to have a permanently visible filter.
+
+For a complete list of the filters available see https://github.com/linzjs/linz/tree/master/lib/formtools/filters.
 
 list.paging
 ===========
--- a/docs/models_list.rst
+++ b/docs/models_list.rst
@@ -122,7 +122,7 @@ list.filters
 
 ``list.filters`` can be used to include filters which will alter the data included in the dataset for a particular model. Filters can contain a custom user interface, but Linz comes with a standard set of filters.
 
-``list.filters`` should be an Object, keyed by each field in your model. Each Object must contain a filter, which should be an object adhering to the Linz model filter API. For example::
+``list.filters`` should be an object, keyed by each field in your model. Each object must contain a filter, which should be an object adhering to the Linz model filter DSL. For example::
 
   filters: {
     dateModified: {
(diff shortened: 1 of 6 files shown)


smebberson:
@allanchau, `'once': true` doesn't quite work as it should. I added it to a filter, and it was removed from the drop down list. I couldn't use it at all.

`once` and `alwaysOn` should 1) be able to work completely independent of each other (they separate features), but 2) should be able to be combined.

So `once` on it's own would have the following affect:

- The filter would appear in the dropdown.
- If you select it, it would be added as an active filter.
- It would have an x next to it.
- It would be greyed out in the dropdown, once active.
- If removed as an active filter, it would be available to choose again.

`alwaysOn` would have the following affect:

- The filter would appear in the dropdown.
- The filter would automatically appear as an active filter.
- It would not have an x next to it.
- You would not be able to remove it.
- If clicked in the drop down, another instance of it would appear.
- That instance would have an x next to it, the other wouldn't.

`once` and `alwaysOn` together would have the following affect:

- The filter would not appear in the dropdown.
- The filter would automatically appear as an active filter.
- It would not have an x next to it.
- You would not be able to remove it.

I hope that explains it a little more?

smebberson:
I think to round off these features from a Linz perspective, we're also going to need the ability to define a default value for `alwaysOn` filters. In the instance of a boolean filter (yes/no radio buttons) you'd definitely want a default to be chosen. In the instance of a multiple-select dropdown list, you might not want a default, which is fine. What do you think?

allanchau:
@smebberson I am not too sure about the defaults. There would be quite a lot of logic needed in determining whether the field is a boolean, text, number etc. We would then need additional logic to determine the field type (radio buttons, category buttons) and these would have to be selected differently?

smebberson:
@allanchau, regarding the defaults, I don't think we need to worry about any of that. That is already taken care of with the current filter functionality.

I think all we'll need to do is force the developers to provide the default value, and then render that to a page using just how Linz renders a pre-populated filter. Then the client-side JavaScript will kick in and it should _just work_.

So we'll need to care of:
1. Updating the DSL to accept a default value.
2. Update Linz where it calls the renderer to provide either the default value, or a modified one (from a form POST).

Do you see something that working?

smebberson:
@allanchau, I'm going to throw this one back for you. I've programmed up the default values and I'd like you to test everything please. If you're happy with it, we can merge it in! And use it!

Please test each type of filter. And test combinations of `alwaysOn`, `once` and `default`, to make sure it works logically. I've changed the client-side JavaScript a little so that you can remove additional `alwaysOn` filters (without `once`) that you might have added, but never remove them all.

Please review the updated documentation here: https://linzjs.readthedocs.io/en/always-on-filters/models_list.html#list-filters

docs/models_list.rst
```diff
@@ -122,15 +122,38 @@ list.filters
...
+
+- ``alwaysOn`` will ensure that the filter is always rendered in the list view.
+- ``default`` allows you to provide a default value for the filter. Useful when using ``alwaysOn``.
+- ``filter`` this is optional, but allows you to specify a filter and should point to a Linz filter, or your own custom one.
+- ``once`` will ensure that a user can only add that filter once (works well with the ``boolean`` filter).
+
+.. note::
+  Be aware of the ``default`` values. Because of Linz's internal query structure most filters will need to provide the ``default`` value as an array, but there are some exceptions.
+
+Below is an example of the ``default`` data type for each filter:
```
allanchau: @smebberson Do you think we can update the code for this to use `[].concat(value)`? That way the user can pass a value or array of values.

PR #146, opened 2017-08-15
#3 - Pagination fix

allanchau:
This PR fixes the pagination when the records to show is greater than the total records. Pagination is now always on unless disabled (all or nothing).


smebberson:
@allanchau, this all looks fine. However, because it contains commits from #143, I won't approve it until after that one is completed and merged in.

PR #159, opened 2017-10-29
Default search

smebberson:
This PR improves Linz model list views, with a default, customisable, keyword search field, that works in-conjunction with filters. It also updates Linz's logic around titles, and simplifies how it works.

There is more in the CHANGELOG.md file. This is a major release, due to breaking changes.

### Setup

- [x] Pull down this branch and configure a project to use it.
- [x] Make sure your models have either a title field in the schema, or provide `model.title`, setting it to the name of a field in your schema that titles should be derived from.

### Testing

- [x] All the steps required to complete testing the PR.
- [x] Test a model with filters and without search.
- [x] Test a model with search and without filters.
- [x] Test a model without search or filters.
- [x] Provide `list.search` with an array of fields it should search, updating the default.
- [x] Test that the new search is all working properly, and works with filters too.
- [x] Test that filters work without search (on models with and without search configured).
- [x] Test the UI at mobile and table viewport sizes.
- [x] Provide `model.labels.title` and make sure it overrides the title field.
- [x] Create a model with a title field, and make sure everything works properly.
- [x] Create a model without a title field, or a value for `model.title`, it should error.
- [x] Set `model.title` to a field that doesn't exist in the schema, it should error.
- [x] Models should now be sorted by date modified, descending, by default.
- [x] Read through the documentation and make sure everything is accurate.


lib/api/model.js
```diff
@@ -62,6 +60,17 @@ function list (user, modelName, callback) {
 }
 
 /**
+ * Retrieve the model DSL for a model
+ * @param  {Object}   user      The user to which the form should be customised
```
allanchau: I think this line can be removed.

public/js/model/index.js
```diff
@@ -28,6 +28,23 @@
 
     });
 
+    $('form.filters').on('submit', function (event) {
```
allanchau: Nothing wrong with this line but I think you are missing a version increment for `linz.css` somewhere.

public/src/css/includes/index.less
```diff
@@ -6,6 +6,10 @@
 
 .model-actions, .options {
 
+  .form-control {
```
allanchau: All these changes are 2 spaced, should they be 4? It's up to you but I prefer 2 spaced without padding for css/less/scss etc anyway.

public/src/css/includes/index.less
```diff
@@ -6,6 +6,10 @@
...
+    height: 42px;
```
allanchau: I'm not too concerned with this since developers don't currently have access to the less files and we are planning on moving to scss (or postcss?) anyway, but this should go in some variable.

allanchau:
@smebberson Can you please update the search placeholder text to be `Search <model name>`? eg `Search Staff Members`. Also I can't seem to disable both filters and search.

lib/api/model.js
smebberson: Done.

public/js/model/index.js
smebberson: Done.

public/src/css/includes/index.less
smebberson: The whole file is 2 spaced, so I kept it consistent.

public/src/css/includes/index.less
smebberson: Yeah, it needs a tidy up. I'll leave it as is for now, because there is no real convention for our CSS yet :(

smebberson:
@allanchau, search placeholder has been updated.

smebberson:
@allanchau, you should now be able to disable both search and filters.

smebberson:
@allanchau, this is ready for another review.

PR #161, opened 2017-10-31
Utilsjs, mongodb options and various updates

### LINKED ISSUE #160: Util.js file
### Expected behaviour

Ability to invoke javascript code using jQuery in a formtool widget.

### Actual behaviour

Javascript using `linz.addLoadEvent();` throws an error now in the latest tag because the `"/admin/public/js/utils.js` has been moved to the footer.

 `linz.addLoadEvent();` was added to allow queueing of javascript code till jQuery is loaded (in the footer)

### Steps to reproduce the behaviour

1. Create a custom widget a jQuery code snippet
2. Run application
3. Check the Chrome Dev Tool, you'll notice an error that says 'Linz' is undefined. This is because the file that instantiate `linz` have been moved to the footer and is not loaded before the widget code.

### LINKED ISSUE #162: linz.api.views.render function not working
### Expected behaviour

Render linz wrapper that provides the navigation and logout buttons etc.

### Actual behaviour

I created a route based on the example https://linzjs.readthedocs.io/en/latest/api.html#views-render-options-callback 

When browsing to that route, it just times out.

### Steps to reproduce the behaviour

1. Use the example code from the link above, hook it up to a route
2. Browse to that route and you'll see it timing out

sandytrinh:
- Fixes #160
- Added new default `mongoOptions` which addresses the new connection configuration introduced in `v4.11.0`+. See [use mongo client](http://mongoosejs.com/docs/connections.html#use-mongo-client) for more info.
- Fixed some syntax issue in `lib/scripts.js`.
- Fixed #162 
- Corrected some code syntax in the documentation

### Setup

- [ ]  Pull down this branch and configure a project to use it.

### Testing

- [ ] View source code of the application and make sure `<script src="/admin/public/js/utils.js"></script>` appears in the html header
- [ ] Test that you can connect to a local mongodb connection URL
- [ ] Test that you can connect to production database connection URL (e.g. mongodb://`user`:`password`@`host`:`port`,`host`:`port`/`db name`?ssl=true
- [ ] Map the example code below to a route and test. You should see a basic page with navigation and logout button with a Header title and Body text.

```
  const linz = require('linz');

  module.exports = function (req, res, next) {

    const locals = {
      header: '<div class="col-xs-12"><div class="model-title"><h1>Header.</h1></div></div>',
      body: '<div class="container linz-container index"><div class="col-xs-12"><p>Body.</p></div></div>'
    };

    linz.api.views.render(locals, req, res, (err, html) => {

      if (err) {
        return next(err);
      }

      return res.send(html);

    });

  };
```

### Notes

Using the new `mongoOptions` I was able to connect to the example URL above without having to load in the SSL certs.

### DIFF
--- a/lib/api/views.js
+++ b/lib/api/views.js
@@ -41,7 +41,7 @@ const viewPath = view => path.resolve(__dirname, '..', '..', 'views', (view || '
  * @param  {Function} callback Optional callback function.
  * @return {String}            Renders and returns the template HTML.
  */
-const render = (data, req, callback) => {
+const render = (data, req, res, callback) => {
 
     // Namespace this request object.
     namespaceMiddleware(req, {}, () => {
@@ -56,8 +56,8 @@ const render = (data, req, callback) => {
                 }
 
                 Promise.all([
-                    getScripts(req, callback),
-                    getStyles(req, callback),
+                    getScripts(req, res),
+                    getStyles(req, res),
                 ])
                     .then(([scripts, styles]) => {
 
--- a/linz.js
+++ b/linz.js
@@ -196,7 +196,7 @@ Linz.prototype.configure = function() {
 
         debugGeneral('Connecting to the database');
 
-        _this.mongoose.connect(_this.get('mongo'));
+        _this.mongoose.connect(_this.get('mongo'), _this.get('mongoOptions'));
         _this.mongoose.connection.once('connected', function () {
             debugGeneral('Database connected');
         });
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -76,7 +76,7 @@ Example
       body: '<div class="container linz-container index"><div class="col-xs-12"><p>Body.</p></div></div>'
     };
 
-    linz.api.views.render(locals, (err, html) => {
+    linz.api.views.render(locals, req, res, (err, html) => {
 
       if (err) {
         return next(err);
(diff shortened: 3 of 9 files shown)


lib/api/views.js
```diff
@@ -41,7 +41,7 @@ const viewPath = view => path.resolve(__dirname, '..', '..', 'views', (view || '
  * @param  {Function} callback Optional callback function.
  * @return {String}            Renders and returns the template HTML.
  */
-const render = (data, req, callback) => {
+const render = (data, req, res, callback) => {
```
allanchau: @smebberson I thought the callback could be a `res` object or `callback` function. Do you remember how it was implemented?

smebberson: @allanchau, you're correct. However, it seems as though `getScripts` and `getStyles` requires `res`. See L59 and L60.

linz.js
```diff
@@ -196,7 +196,7 @@ Linz.prototype.configure = function() {
 
         debugGeneral('Connecting to the database');
 
-        _this.mongoose.connect(_this.get('mongo'));
+        _this.mongoose.connect(_this.get('mongo'), _this.get('mongoOptions'));
```
allanchau: I think we need some sort of `Object.assign({}, _this.get('mongoOptions'), the options from the passed in mongoose)` here. Does Linz already do something like that @smebberson ?

smebberson: @allanchau, so the `_this.get('mongo')` and `_this.get('mongoOptions')` just pulls values from the defaults, unless they've been overrided at init time. What benefit would `Object.assign` have here?

smebberson: @allanchau, also, just note, this won't get called if the passed in Mongoose is already connected.

lib/api/views.js
smebberson: I think we should keep this change, so that it requires `res`, but update the code further below so that callback is an optional argument. It won't take much.

linz.js
allanchau: Ah ok, this is all good then.

smebberson: 🎉

PR #167, opened 2017-11-06
List count fix

smebberson:
This PR fixes an issue in which the record count reported on a model list, could be incorrect if the model has been further filtered by a `listQuery` static.

This PR resolves that.

### Setup

- [x] Pull down this branch.
- [x] Wire it up into your project.

### Testing

- [x] Test that record count is accurate, on a model without `listQuery`.
- [x] Add a `listQuery` to your model that will always ensure some records aren't shown. The record count should remain accurate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## Unrelease
+
+### IMPROVEMENTS
+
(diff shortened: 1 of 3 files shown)


allanchau:
@smebberson All good, I'll leave it with you to merge this and the other PR + do a release.

PR #170, opened 2017-11-09
Model list IE related fixes

smebberson:
Fixes a problem with the model list in IE. Due to the use of templates, IE was having a hard time with it.

When hitting search, with a value in the search field, everything was okay.
When hitting search with no value in the search field, it didn't work.

### Setup

- [x] Pull down this branch of Linz.
- [x] Configure your project to use it.

### Testing

- [x] Make sure you can use the search field, without any filters.
- [x] Make sure you can use the filters, without the search field.
- [x] Make sure you can use both together.
- [x] Test this ie IE, Edge, Safari, Firefox and Chrome.


routes/recordOverview.js
```diff
@@ -13,6 +13,9 @@ var route = function (req, res, next) {
             {
                 src: `${linz.get('admin path')}/public/js/model/index.js?v3`,
             },
+            {
+                src: `${linz.get('admin path')}/public/js/template.pollyfill.js`,
```
allanchau: @smebberson Isn't it `polyfill`? :)

allanchau:
@smebberson All good, can you please do a release.

PR #171, opened 2017-11-11
Replace user with req

smebberson:
This PR will add more flexibility in what you can do with the model form, list and overview DSLs. Before this PR, when using a function, you could only base the result on the user making the request. This PR allows you to base the result on the record itself and the user making the request as these functions are now passed `req`, rather than `user`.

This is the first in a new direction where all future APIs will involved passing `req` where possible.

### Setup

- [ ] Pull down this branch, and map it into your project.

### Testing

- [ ] Before updating your project, your list, form and overview views should either break or not work properly if you're providing a function.
- [ ] Update your functions to accept `req` rather than `user`.
- [ ] Check that the model form function receives `req`, and can be rendered properly.
- [ ] Check that the model list function receives `req` and can be rendered properly.
- [ ] Check that the model overview function receives `req`, and can be rendered properly.
- [ ] Check that the config form function receives `req`, and can be rendered properly.
- [ ] Check that the config list function receives `req`, and can be rendered properly.
- [ ] Check that model creation works.
- [ ] Check that record editing works.
- [ ] Check that record deletion works.
- [ ] Check that config editing works.
- [ ] Read the documentation and make sure it's all okay.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,13 @@
 ### BREAKING CHANGES
 
 - When rendering the overview, Linz will no longer infer the cell renderer type from the Form DSL type data. If Linz isn't rendering the overview cell with the correct cell renderer, you should supply the renderer (i.e. `{ label: 'x', renderer: linz.formtools.cellRenderers.text }`).
+- When supplying a function for the Model overview DSL, you no longer get passed `user`, but instead `req` (which obvious has user at `req.user`).
 
 ### IMPROVEMENTS
 
 - Fixes `#169`, an issue in which the actions column wouldn't render when permissions weren't explicitly set.
 - Improved the Express param for model. It no longer process list, overview or form DSL. These are now only processed when required. Improving the efficiency of any route that used `:model`.
+- When supplying functions for Model overview DSL, you now get `req` rather than just `user`. This provides much more flexibility in what you return, as now it can be based on the record being rendered, not just the user making the request.
 
 ## v1.0.0-13.0.2 (9 November 2017)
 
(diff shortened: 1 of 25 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
 # CHANGELOG
 
+## Unreleased
+
+### BREAKING CHANGES
+
+- When rendering the overview, Linz will no longer infer the cell renderer type from the Form DSL type data. If Linz isn't rendering the overview cell with the correct cell renderer, you should supply the renderer (i.e. `{ label: 'x', renderer: linz.formtools.cellRenderers.text }`).
+- When supplying a function for the model list, form and overview DSLs, you no longer get passed `user`, but instead `req` (which obviously has user at `req.user`).
+- When supplying a function for the config form and overview DSLs, you no longer get passed `user`, but instead `req` (which obviously has user at `req.user`).
```
allanchau: config form can just be added to the line above?

CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
...
+
+### IMPROVEMENTS
+
+- Fixes `#169`, an issue in which the actions column wouldn't render when permissions weren't explicitly set.
+- Improved the Express param for model. It no longer process list, overview or form DSL. These are now only processed when required. Improving the efficiency of any route that used `:model`.
+- When supplying functions for model list, form and overview DSLs, you now get `req` rather than just `user`. This provides much more flexibility in what you return, as now it can be based on the record being rendered, not just the user making the request.
+- When supplying functions for config form and overview DSLs, you now get `req` rather than just `user`. This provides much more flexibility in what you return, as now it can be based on the record being rendered, not just the user making the request.
```
allanchau: Same here?

lib/api/configs.js
```diff
@@ -52,24 +52,24 @@ function permissions (user, configName, callback) {
...
-function form (user, configName, callback) {
-    return get(configName).schema.statics.getForm(user, callback);
+function form (req, configName, callback) {
+    return get(configName).schema.statics.getForm(req, callback);
 }
 
 /**
  * Retrieve the overview DSL for a config
- * @param  {Object}   user      The user to which the overview DSL should be customised
+ * @param  {Object}   req      A HTTP request object which should be used to customise the DSL
```
allanchau: Just for consistency this should read `can be used...`

PR #180, opened 2017-12-08
Added support for scripts and styles to have a content property

smebberson:
Scripts and styles can now have a content property.

### Setup

- [x] Pull down this branch, and map it into your project.

### Testing

- [x] Add a `content` property to a script. Check content is output within `script` tag.
- [x] Add a `content` property to a script that also has `inHead`. Check content is output within `script` tag.
- [x] Add a `content` property to a style. Check content is output within `style` tag.
- [x] Add a `content` property to a style that also has `inHead`. Check it is ignored.
- [x] Read documentation and make sure it all makes sense.

### Notes

- This is useful for adding things like Google Tag Manager and other scripts which require being within a script tag.

### DIFF
--- a/views/partials/foot.jade
+++ b/views/partials/foot.jade
@@ -1,7 +1,10 @@
 if scripts && scripts.length
 	each script in scripts
 		if !script.inHead
-			script(async=script.async crossorigin=script.crossorigin defer=script.defer integrity=script.integrity src=script.src text=script.text type=script.type)
+			if script.content
+					script(async=script.async crossorigin=script.crossorigin defer=script.defer integrity=script.integrity src=script.src text=script.text type=script.type)!= script.content
+				else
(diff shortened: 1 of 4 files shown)


views/partials/foot.jade
```diff
@@ -1,7 +1,10 @@
 if scripts && scripts.length
 	each script in scripts
 		if !script.inHead
-			script(async=script.async crossorigin=script.crossorigin defer=script.defer integrity=script.integrity src=script.src text=script.text type=script.type)
+			if script.content
+					script(async=script.async crossorigin=script.crossorigin defer=script.defer integrity=script.integrity src=script.src text=script.text type=script.type)!= script.content
+				else
```
allanchau: I think this block is indented too much?

smebberson: It just looks large here. It's using tab indentation. Looks fine in the editor, and it is inline with the script tag on L7. Looks fine to me...

PR #185, opened 2017-12-14
ON HOLD: Query plugin

allanchau:
## Update

I think we should hold off on merging this in until https://github.com/linzjs/linz/issues/193 is done. We can then create a new repo for it and use it there.

This PR adds a new global query plugin that adds the static functions `findOneDocument` and `findDocuments` to all models.

### Setup

- [ ] Map in this version of Linz.

### Testing

- [ ] Update your model queries to use `findOneDocument` and `findDocuments`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,15 @@
 # CHANGELOG
 
+## Unreleased
+
+### Improvements
+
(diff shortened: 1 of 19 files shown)


allanchau:
This is probably something better handled in a plugin when that eventually comes to Linz.

PR #187, opened 2017-12-19
Form api

allanchau:
This PR adds the ability to create a Linz form using the form DSL.

**Testing**

- [x] Click on the record action for a `mtUser` record.
- [x] Make sure the existing fields are pre-populated.
- [x] Submit the form and make sure all fields in the form are passed.
- [x] Make sure child schemas work in the model form and custom form.
- [x] Make sure the values in the custom form post route are correct. They should be of the same type that was specified `number: '10' should become number: 10`.
- [x] Make sure the datetime input works on modals.
- [x] Add this PR to a large project to test all functionality (including config items)

**Notes**

- Have a review of the updated devops functionality, it turns out `--link` is deprecated and creating a network should be used instead.


lib/api/middleware/set-linz-model.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+const linz = require('../../../');
+
+/**
+ * Set req.linz.model
+ * @returns {Void} Calls the next middleware.
+ */
+const setLinzModel = () => (req, res, next) => {
```
smebberson: Why does this need to be a function that returns a function? There are no arguments, so it would seem unnecessary?

lib/api/middleware/set-model-form.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+const setLinzNamespace = require('./set-linz-namespace');
+
+/**
+ * Set req.linz.model.form.
+ * @returns {Void} Calls the next middleware.
+ */
+const setModelForm = () => (req, res, next) => {
```
smebberson: Why does this need to be a function that returns a function? There are no arguments, so it would seem unnecessary?

lib/api/middleware/set-record.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+const setLinzNamespace = require('./set-linz-namespace');
+
+/**
+ * Set req.linz.record.
+ * @returns {Void} Calls the next middleware.
+ */
+const setRecord = () => (req, res, next) => {
```
smebberson: Why does this need to be a function that returns a function? There are no arguments, so it would seem unnecessary?

lib/defaults.js
```diff
@@ -44,13 +44,13 @@ module.exports = {
 
     // sets an array of middleware functions to handle login
     'login middleware': {
-        get: [require('../middleware-public/login').get],
-        post: [require('../middleware-public/authenticate')('linz-local', { failureFlash: true })]
+        get: [require('./api/middleware/login').get],
```
smebberson: I think it would be best to move the `require` statements to the top of this file?

allanchau: I've found moving modules to the top can cause circular dependencies where the module becomes `{}`. I haven't actually tried it here, so it might not change anything.
Did you want me to update it?

allanchau:
@smebberson I think this is ready for another round of reviews.

allanchau:
@smebberson This is ready for you now.

lib/defaults.js
smebberson: @allanchau, let's try moving it to the top.

lib/formtools/form.js
```diff
@@ -225,14 +232,11 @@ var generateFormFromModel = exports.generateFormFromModel = function (m, schemaF
...
-                    switch (field.type) {
+                    switch (fieldType) {
 
                         case 'date':
-                        fieldValue = ((r[fieldName] !== undefined && r[fieldName] !== null) ? moment(r[fieldName]).format('YYYY-MM-DD') : null) || defaultValue;
-                        break;
-
                         case 'datetime':
-                        fieldValue = ((r[fieldName] !== undefined && r[fieldName] !== null && (r[fieldName]) instanceof Date) ? r[fieldName].toISOString().slice(0,23) : defaultValue) || defaultValue;
+                            fieldValue = r[fieldName] !== undefined && r[fieldName] !== null ? moment(r[fieldName]).toISOString() : defaultValue;
```
smebberson: Have you tested for this change specifically (i.e. `date` and `datetime`)?

lib/formtools/model.js
```diff
@@ -1,8 +1,36 @@
...
+ * @param {Object} body The request body to clean.
+ * @param {Object} model The model.
+ * @return {Object} The cleaned request body.
+ */
+const clean = (body, model) => {
+
+    // Don't modify the original.
+    const data = Object.assign({}, body);
+
+    if (!body) {
```
smebberson: Shouldn't this come before the code above?

docs/api.rst
```diff
@@ -12,6 +12,53 @@ At present, we're working on a long term effort to expose all of Linz's function
...
+model: object
++++++++++++++++
+
+The mongoose model.
+
+options: object
++++++++++++++++
+
+========== ====== ===========
+Property   Type   Description
```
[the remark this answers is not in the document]
allanchau: Nope, there are defaults for all the options.

lib/formtools/form.js
allanchau: Yeah, there is a bit of a problem here though.
We have been updating the date widget, but not the datetime or datetimelocal widget. I think neither of these have worked for a while.

Did you want me to fix those up or remove them completely from Linz? I don't see much point in keeping them since the date widget can handle everything that these other widgets can.

lib/formtools/utils.js
```diff
@@ -35,7 +35,7 @@ exports.getDefaultValue = function (field) {
             return value = field.defaultValue();
         }
 
-    } else if (field.defaultValue === undefined && fieldType === 'datetime') {
+    } else if (field.defaultValue === undefined) {
```
smebberson: Do you think this is appropriate without a check against the `fieldType`? It means if there is no default value, it will set the value to a date? That doesn't seem quite right to me?

smebberson:
@allanchau, I think it's a mistake that we (looks like me) updated this so that you can no longer run async code within the formDSL function (that was the whole point of the function!!). Could you please update it to accept a callback again?

lib/formtools/utils.js
smebberson: @allanchau, this one is still to be considered.

allanchau: @smebberson Should this block be removed altogether? So the default is null if not type is specified.

smebberson: @allanchau, that's what I'm thinking... As long as the default fields that are applied by Linz include a default, I think it should be fine. Do you agree?

allanchau: Yeah, sounds good to me.

PR #188, opened 2018-01-16
Loadevent fix

allanchau:
The `linz.addLoadEvent` can be overridden when using the model helpText. This PR updates the code so that `linz.addLoadEvent` will append to `window.onload` and not replace it.

### Testing

- [x] Test out the helptext in a model, it should break.
- [x] Map in this branch and reload the page, the helptext should be working again.


smebberson:
@allanchau, can you please browser check in Chrome, Firefox, Safari and IE 11 and IE Edge.

allanchau:
@smebberson It works in all the major browsers for Windows and Mac based on my testing.

PR #189, opened 2018-01-19
Empty actions fix

allanchau:
This PR prevents linz from displaying empty add filter dropdowns.

### Setup

- [x] Create a model with a single alwaysOn filter.

### Testing

- [x] Verify there is no filter dropdown.


smebberson:
@allanchau, can you please add a cache miss for the JavaScript file...

allanchau:
@smebberson done

PR #197, opened 2018-03-12
Fix the count for index views

allanchau:
This PR fixes the list view count when listQuery filters on the results.

### DIFF
--- a/middleware/_modelIndex.js
+++ b/middleware/_modelIndex.js
@@ -283,6 +283,35 @@ module.exports = function  (req, res, next) {
 
             },
 
+            // minimise the fields we're selecting
+            function (cb) {
+
+                let fields = Object.keys(req.linz.model.list.fields);
+
+                // Work in the title field
+                linz.api.model.titleField(req.params.model, 'title', (err, titleField) => {
+
+                    if (err) {
+                        return cb(err);
+                    }
+
+                    fields.push(titleField);
+
+                    const select = fields.join(' ');
+
+                    query.select(select);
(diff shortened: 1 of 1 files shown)


smebberson:
@allanchau, this one is a confusing one. I only just recently made commit https://github.com/linzjs/linz/commit/fd99e561e50ebfec41986367de41fc1024e72e38 which was put in place to resolve https://github.com/aheckmann/mquery/issues/55. I'm not exactly sure how it happened, but all of a sudden any of my Linz projects where throwing a strange error relating to that. Moving the count above `listQuery` resolved it for me.

Perhaps we should split these up into two parts? Have the `listQuery` called first, then call `count()` then do the `select` and `sort` parts?

allanchau:
@smebberson Do we even need the getCount static method? We could easily do a `totalRecords = docs.length` at ~L392 and remove a db query?

smebberson:
@allanchau, the `getCount` static method is ultra important isn't it? It provides information about total rows for the current query and powers things such as paging in which knowing the total results is imperative. How would it work without it?

allanchau:
@smebberson From what I can see there are currently 3 queries for the index pages:
1. Get the docs
2. Get the total records
3. Get the pageIndex (not really sure how `session.list.formData.page` is set)

Only the total records seem to use `req.linz.model.getCount`?
You would still get the total records using `docs.length` instead of creating another query.

smebberson:
@allanchau, after our chat and the test, this looks good.

PR #198, opened 2018-03-16
Mongoose update

allanchau:
This PR updates linz to fix a kerberos error (nogodb native driver uses it) when running npm install.

### Testing

- [ ] Run npm install and make sure there is no kerberos error.


allanchau:
Doesn't fix it 😢

PR #201, opened 2018-03-20
Fix templates for ie

allanchau:
This PR fixes the template tag for ie.

Signed-off-by: Allan Chau <allan.chau@icloud.com>

## Related PRs

- #188 

## Setup

- [x] Test an alwaysOn filter in IE and verify it breaks.

## Testing

- [x] Pull in this PR and verify the filter now works.

## Notes

- The related pr should be loaded in first as it requires the `addLoadEvent` function.


smebberson:
@allanchau, you forgot the cache miss here...

allanchau:
@smebberson It was updated in the previous merge? Did you want me to update it again?

smebberson:
I think it should be updated each and every time a public JS file is changed. Yes, please update.

allanchau:
@smebberson done now

PR #204, opened 2018-03-25
Only default to descending with date fields

allanchau:
This PR adds the ability to provide an object or string for the sortby array.

Signed-off-by: Allan Chau <allan.chau@icloud.com>

### Setup

- [x] Create a model with a String and object in the sortby array. 

```javascript
module.exports = [
    'title',
    'description',
    {
        defaultOrder: 'desc', // Chronological order
        field: 'dateModified',
        label: 'Date modified',
    },
];
```

### Testing

- [ ] Using the example above, sort by the title. When the current sorting is **NOT** the default there should be no `descending` or `ascending` text.
- [ ] Filter on the dateModified field. It should default to descending.
- [ ] View the sorting list and it should say `Date modfied (ascending)`

### DIFF
--- a/lib/formtools/plugins/plugins-helpers.js
+++ b/lib/formtools/plugins/plugins-helpers.js
@@ -533,10 +533,18 @@ var transformSortBy = function (labels, sortBy) {
 
     return sortBy.map(function (fieldName) {
 
-        return {
+        const defaults = {
+            defaultOrder: 'asc',
             label: (labels[fieldName] || fieldName),
-            field: fieldName
+            field: fieldName,
         };
+
+        // Add the ability to pass an object.
+        if (typeof fieldName !== 'string') {
+            return Object.assign({}, defaults, fieldName);
+        }
+
+        return defaults;
 
     });
 
--- a/lib/formtools/plugins/plugins-helpers.js
+++ b/lib/formtools/plugins/plugins-helpers.js
@@ -68,7 +68,10 @@ var listDefaults = function (list, labels, titleField) {
     // defaults to true
     list.showSummary = !(list.showSummary === false);
 
-    list.sortBy = transformSortBy(labels, list.sortBy || ['dateModified']);
+    list.sortBy = transformSortBy(labels, list.sortBy || [{
+        defaultOrder: 'desc',
+        field: 'dateModified',
+    }]);
 
     // Will accept an object of Bootstrap popover options.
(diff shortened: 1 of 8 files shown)


smebberson:
@allanchau, can you please add some documentation for this. Also, it seems as though this is a breaking change? Did you notice the builds are failing?

smebberson:
I re-read the code. It's not a breaking change.

allanchau:
Yeah sorry, I didn't realise they were failing. I'll update the tests in a bit.

lib/formtools/plugins/plugins-helpers.js
```diff
@@ -533,9 +536,25 @@ var transformSortBy = function (labels, sortBy) {
...
+        // Add the ability to pass an object.
+        if (typeof fieldName !== 'string') {
+
+            if (!fieldName.field) {
+                throw new Error('You must provide a field');
+            }
+
+            return Object.assign({}, {
+                defaultOrder: 'asc',
+                label: (labels[fieldName.field] || fieldName.label),
```
smebberson: What if they haven't provided a label? It doesn't seem to default the label to the `fieldName` itself, if the field doesn't exist in the `label` array and the `label` wasn't provided in the `sortBy` object?

allanchau: @smebberson So should we just add another fallback `|| fieldName.field`?

smebberson: @allanchau, in fact, it already defaults when the label is missing. I think that is because the label array is defaulted with the `fieldname` if a label isn't provided. So I don't think it's necessary - you can leave it as is.

smebberson:
@allanchau, something isn't working with defaultOrder field. I had the following setup:

```
    sortBy: [
            'dateModified',
            {
                defaultOrder: 'desc',
                field: 'username',
                label: 'Userrmame',
            },
        ]
```

When the page was rendered, date modified was the default, in ascending order which is correct. I then chose `Userrmame` from the drop-down list, however, it was rendered in ascending order, but I thought it should have been descending order.

smebberson:
@allanchau, I've taken the opportunity to change something in Linz I've always wanted too, which is, it now includes both ascending and descending options (for one field) in the drop down itself. This will increase usability as you often had to select a field, then choose descending causing more clicks, more requests, etc.

PR #211, opened 2018-07-31
Login override

allanchau:
This PR adds the ability to override the login and logout paths.

**Setup**

- [x] Add the following option top the linz init method `'login path': '/gh-211/login',`
- [x] Add the following route under the default root route`linz.app.get('/gh-211/login', (req, res) => res.send('gh-211'));`

**Testing**

- [x] Access `localhost:8888`, it should redirect you to the new path.

**Notes**

- This is intended to allow custom login pages, the new path will not use the linz login middleware by default.


smebberson:
@allanchau, tests aren't passing. Could you please take a look?

allanchau:
@smebberson I've fixed the tests now.

PR #218, opened 2018-08-07
Ref link

allanchau:
This PR updates the list reference renderer to link to the ref record.

## Setup

- [x] Visit `localhost:8888` to get the updated models.

## Testing

- [x] View the user model.
- [x] Make sure the ref field links to the organisation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 # CHANGELOG
 
-## Unreleased
+## v1.0.0-16.1.0 (9 August 2018)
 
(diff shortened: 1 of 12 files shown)


allanchau:
I've updated this to include more tests. Unfortunately I also had to split up the `loadModels` function to be able to test the reference renderer since Linz didn't set up the models completely unless was loaded via file.

allanchau:
@smebberson Are you able to approve + merge this in again? I just pushed an update to the readme build badge.

PR #224, opened 2018-11-20
Export functionality show allow customising how the fields are displayed.

### LINKED ISSUE #223: Export functionality show allow customising how the fields are displayed.
**Is your feature request related to a problem? Please describe.**
Currently there is no way to customise the selected fields before exporting it to a file.

**Describe the solution you'd like**
Each field should be customisable before outputting to a file.

**Additional context**
The export functionality should use the existing `transpose` property in the formDSL.
This should be updated to have contexts such as `form` and `export`. If these are not supplied, the passed function will be used for all contexts.
The passed function should also take the same parameters as the cell renderers so that these can be used to transpose the field.

allanchau:
This PR allows customising how fields should display in the default export functionality.

**Setup**

- [x] Add this to a project that uses the export functionality in Linz.

**Tasks**

- Fixes #223 

**Testing**

- [x] The mtUser model has been updated to transpose the name field. Make sure it replaces the name field with ``Exporting ${val} name`` when exporting and ``Transposing the ${val} using the form context`` when editing the record.
```javascript
transpose: {
    'export': (val) => Promise.resolve(`Exporting ${val} name`),
    'form': (val) => Promise.resolve(`Transposing the ${val} using the form context`),
},
```
- [x] Replace the transpose property with the below. Make sure the field is not transposed when exporting. Viewing the record should transpose the field as per before, but it should now show a deprecation notice in the console.
```javascript
 transpose: (val) => `General export ${val}`
```


app/models/mtUser.js
```diff
@@ -88,6 +90,13 @@ mtUserSchema.plugin(linz.formtools.plugins.document, {
                 label: 'Customised edit form',
             },
         ],
+        export: [
+            {
+                exclusions: '_id',
+                label: 'Choose fields to export',
+                dateFormat: 'DD MMM YYYY',
```
smebberson: @allanchau, what does this line do?

lib/api/util/file-export.js
```diff
@@ -0,0 +1,57 @@
+'use strict';
+
+const csvTransform = require('stream-transform');
+const csvStringify= require('csv-stringify');
```
smebberson: @allanchau, should be a space before the `=` here.

lib/api/util/index.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+const fileExport = require('./file-export');
+
+module.exports = Object.assign({}, fileExport);
```
smebberson: @allanchau, why the `Object.assign` here?

app/models/mtUser.js
allanchau: There is code in Linz that allows you to format dates using `dateFormat`. 🤷‍♂️ I don't remember when it was added, but it does seem to work.

middleware/recordChange.js
```diff
@@ -36,7 +36,7 @@ module.exports = function (req, res, next) {
 			}
 
 			// if transpose is defined for this field, let's tranpose their change so it can compare in the correct format on client side
-			if (form[fieldName].transpose) {
+			if (form[fieldName].transpose && typeof form[fieldName].transpose === 'function') {
```
smebberson: @allanchau, I think we need to extend this. At present, if you have `transpose.export`, you can't have a different transpose function specific to the `form` use?

I think we'll need to update this to have `transpose.export` and `transpose.form`... what do you think?

lib/api/util/index.js
allanchau: Just planning ahead for when we add more functions here. We don't need it now though, did you want me to remove it?

smebberson: @allanchau, of course, yeah, that's better and more inline with our standard coding...

middleware/recordChange.js
allanchau: yeah ok, sounds good.

smebberson: @allanchau, cool, thanks. Please update the documentation while you're at it outlining the different contexts... thanks.

smebberson: Maybe could write it like the following, so it's not a major change now (and we can deprecate it later?):

```
if (form[fieldName].transpose && typeof form[fieldName].transpose === 'function' || form[fieldName].transpose && form[fieldName].transpose.form && typeof form[fieldName].transpose.form === 'function')
```

smebberson: Well, if we're deprecating, not in one like that, because we'll need to do the deprecating stuff...

smebberson: i.e. https://www.npmjs.com/package/deprecate

allanchau:
@smebberson I've updated it now. Instead of using that package, I've just used the built in node util deprecation function. Note the updated testing steps.

CHANGELOG.md
```diff
@@ -10,6 +15,9 @@
 
 ### IMPROVEMENTS
 
+- Updated the modelExport route to `transpose` fields if a `transpose` or `transpose.export` function is provided in the formDSL. This allows you to completely customise exported fields before displaying them in the exported file.
```
smebberson: @allanchau, are these in the wrong spot?

docs/models_form.rst
```diff
@@ -278,10 +278,31 @@ In some instances, client side data requirements are different from that of data
...
+  transpose.form (field, record)
 
 If provided, it will be executed before a field's value is rendered to a form. It is useful if you'd like to manipulate the server-side data that is rendered to a form.
 
 In some instances, data storage requirements are different form that of client side data requirements. ``transpose`` in combination with ``transform`` can be used effectively to manage these scenarios.
+
+{field-name}.transpose.export
+======================
+
+The ``transpose.export`` property can accept a function with the signature::
```
smebberson: I think this should also get `record`. While not useful for the default export, it could be useful when exporting in other scenarios?

middleware/recordChange.js
```diff
@@ -33,11 +38,19 @@ module.exports = function (req, res, next) {
...
 				return;
-			}
+            }
+
+			// if transpose is defined for this field, let's transpose their change so it can compare in the correct format on client side
+			if (getTransposeFn(form, fieldName, 'form')) {
+
+                let transposeFn = getTransposeFn(form, fieldName, 'form')(theirChange[fieldName], theirChange);
+
+                if (!(transposeFn instanceof Promise)) {
```
smebberson: Is this necessary? The Promise support is specific to the contexts `form` and `export` and neither of them have ever supported anything then Promise so I don't think the deprecating is a good thing here? I know the `transpose` itself used to support a function and that we're deprecating that. I don't think the transpose function itself should support Promises. Does this make sense?

allanchau: @smebberson This line has to be wrapped in a promise in order to support the `promises.push` below. `getTransposeFn` is returning the transpose function in the form dsl and this is checking to make sure it is a function (deprecated) or a Promise?

PR #232, opened 2019-02-13
Add CSRF protection to all Linz forms

allanchau:
This PR enables csrf protection across all Linz routes.

**Testing**

- [x] Test the admin login.
- [x] Test the model index page.
- [x] Test a filter on a model index page.
- [x] Test the overview page.
- [x] Test a record action.
- [x] Test editing records.
- [x] Test creating records.
- [x] Test deleting records.
- [x] Test a custom form.

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -48,3 +48,10 @@ When implementing a custom page with a form, make sure to pass the following hid
 This is handled automatically for you when using ``linz.api.model.generateFormString()``. Just make sure to add the `csrfToken` option::
 
     linz.api.model.generateFormString(linz.api.model.get('user'), { csrfToken: req.csrfToken() });
+
+Adding CSRF protection to a non linz route
+==========================================
+
+Linz exposes a csrf api ``linz.api.middleware.csrf`` which can be used to provide CSRF protection for a non Linz route, for example an api route might look something like this::
+
+    linz.app.use('/api', linz.api.middleware.csrf, apiRoute);
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -48,10 +48,3 @@ When implementing a custom page with a form, make sure to pass the following hid
 This is handled automatically for you when using ``linz.api.model.generateFormString()``. Just make sure to add the `csrfToken` option::
 
     linz.api.model.generateFormString(linz.api.model.get('user'), { csrfToken: req.csrfToken() });
-
-Adding CSRF protection to a non linz route
-==========================================
-
-Linz exposes a csrf api ``linz.api.middleware.csrf`` which can be used to provide CSRF protection for a non Linz route, for example an api route might look something like this::
-
-    linz.app.use('/api', linz.api.middleware.csrf, apiRoute);
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -1,0 +1,50 @@
+.. highlight:: javascript
+
+.. _security-reference:
+
+********
+Security
+********
+
+Linz comes with a number of security features out of the box including:
+
+- CSRF Protection
+
+CSRF Protection
+----------------
+
+CSRF protection helps prevent unauthorized commands that are transmitted from a user that the web application trusts.
+
+You can read more about CSRF here https://en.wikipedia.org/wiki/Cross-site_request_forgery.
+
+Customising CSRF
+================
+
+Linz uses the csurf module to provide CSRF protection. https://github.com/expressjs/csurf
+
+To customise the options you can supply Linz with the option ``'csrf options: {}'``. It accepts an object with the same properties as the csurf module::
+
+    linz.init({
+        'options': {
+            'csrf options': {},
+        },
+    });
+
+
+Custom error handler
+====================
+
+CSRF errors throw an error with the code ``err.code === 'EBADCSRFTOKEN'``. You can use this in your error handlers to display a custom message.
+
+Adding CSRF protection to a custom form
+============================================
+
+The csurf module exposes the csrf token via ``req.csrfToken()``.
+
+When implementing a custom page with a form, make sure to pass the following hidden input::
+
+    <input type="hidden" name="_csrf" value="{{csrfToken}}">
+
+This is handled automatically for you when using ``linz.api.model.generateFormString()``. Just make sure to add the `csrfToken` option::
+
+    linz.api.model.generateFormString(linz.api.model.get('user'), { csrfToken: req.csrfToken() });
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -15,12 +15,12 @@ CSRF Protection
 
 CSRF protection helps prevent unauthorized commands that are transmitted from a user that the web application trusts.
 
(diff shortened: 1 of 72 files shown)


docs/security.rst
```diff
@@ -0,0 +1,50 @@
...
+Linz comes with a number of security features out of the box including:
+
+- CSRF Protection
+
+CSRF Protection
+----------------
+
+CSRF protection helps prevent unauthorized commands that are transmitted from a user that the web application trusts.
+
+You can read more about CSRF here https://en.wikipedia.org/wiki/Cross-site_request_forgery.
```
smebberson: Could you please update the link here to https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF), as they're seen as the industry standard when it comes to security.

docs/security.rst
```diff
@@ -0,0 +1,50 @@
...
+
+Customising CSRF
+================
+
+Linz uses the csurf module to provide CSRF protection. https://github.com/expressjs/csurf
```
smebberson: Could you please link the text _csurf module_ to the link you provided using the RST external link syntax... http://docutils.sourceforge.net/docs/user/rst/quickref.html#external-hyperlink-targets

docs/security.rst
smebberson: Could you please replace `more about CSRF here` with `more about CSRF on OWASP` and make the link text _CSRF on OWASP_ using the RST external link syntax... http://docutils.sourceforge.net/docs/user/rst/quickref.html#external-hyperlink-targets

docs/security.rst
```diff
@@ -0,0 +1,50 @@
...
+        'options': {
+            'csrf options': {},
+        },
+    });
+
+
+Custom error handler
+====================
+
+CSRF errors throw an error with the code ``err.code === 'EBADCSRFTOKEN'``. You can use this in your error handlers to display a custom message.
```
smebberson: @allanchau, could you please throw in a quick example here...

lib/csrf.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+const csurf = require('csurf');
+
+module.exports = (linz) => csurf(linz.get('csrf options'));
```
smebberson: Can't this retrieve `linz` itself? It's not a pattern we've really used before...

middleware/modelExport.js
```diff
@@ -261,11 +261,11 @@ module.exports = {
             req.linz.export = getExport(list.export);
             req.linz.export.fields = {};
 
-            // retrieve the form to provide a list of fields to choose from
-            req.linz.model.getForm(req, function (formErr, form){
+            // retrieve the labels to provide a list of fields to choose from
+            req.linz.model.getLabels((labelErr, labels) => {
```
smebberson: @allanchau, should we update this to use the model schema rather than the labels? The labels might not be complete either...

middleware/modelExport.js
```diff
@@ -329,7 +329,11 @@ module.exports = {
 
         });
 
-        async.waterfall(asyncFn, function (err, filters, form, list, exportObj) {
+        asyncFn.push(function (filters, form, list, exportObj, cb) {
```
smebberson: @allanchau, you can't see it in this diff, but it looks as though the previous asyncFn function returns `null` as the first callback, which I'm guessing is a space to hold an `err` as per Node.js convention. However your function doesn't provide a spot for it but instead would make `filters` null?

smebberson: Am I missing something?

package.json
```diff
@@ -55,7 +56,7 @@
     "less-middleware": "1.0.3",
     "moment": "2.22.2",
     "mongoose": "4.13.12",
-    "mongoose-version": "github:flashmedia/mongoose-version",
+    "mongoose-version": "idearium/mongoose-version#4ec7aa5b6bd628a7da8e7ab4ce1888bb74ff8b6b",
```
smebberson: @allanchau, we should actually move this to the `linzjs` organisation?

smebberson: It looks like the build on that commit is failing?

lib/csrf.js
allanchau: @smebberson Do you mean require linz in the function?
I tried just using `linz.get`, but I don't think linz is initialised at that point so we don't have the `get` function.

middleware/modelExport.js
allanchau: In a previous update, the defaults were added to all the labels.
If we used the schema, I'm not sure how we could grab the label?

middleware/modelExport.js
allanchau: hrmm in the next line I am passing `err` to the callback (which should be undefined).
It is pretty much the same as the block above L322 except it adds labels to the parameters?

package.json
allanchau: @smebberson yeah, I fixed a syntax error `'use strict'` in that commit, but I didn't want to change how the code works. I don't think we can fix the tests without doing that because they currently run on an extremely outdated node version 0.X 😱

lib/csrf.js
smebberson: Yeah, I mean using the requiring Linz. So you tried that and it didn't work?

middleware/modelExport.js
smebberson: Right so the purpose here is to grab the labels, rather than a list of all fields?

middleware/modelExport.js
smebberson: But that `err` is coming from the `getLabels` call...

allanchau: Yeah so if there is an error, err will be defined and the asyncFn will go straight to L339.
If there is no error, it will skip L338 because err is undefined.

The previous async function just says it will never fail because err is always null.

middleware/modelExport.js
allanchau: Sort of, it is used to display the fields in the popup and as the csv headers.
`L63 /lib/formtools/plugins/document.js` seems to provide complete coverage for all fields besides _id?

middleware/modelExport.js
smebberson: Yeah right, I get you. Makes sense. Thanks.

lib/csrf.js
allanchau: Yeah, the get function wasn't defined yet. Without restructuring the linz initialisation code, I don't think we can just require linz.

smebberson: If you move the `require('./lib/csrf');` in `linz.js` down a little until after Linz has initialised it will work. I think that's the lesser of the two evils at the moment and in more line with other Linz coding?

package.json
smebberson: @allanchau, all builds are failing, even those before your updates. Do we even need to version bump the Mongoose dependency? Does that specifically fix anything?

allanchau: It fixes a warning during startup, since we are using it on mongoose 4 already I though it would be fine.

I couldn't get it to build without the `lib/clone-schema.js` change though. At a minimum it needs that.

smebberson:
@allanchau, this looks good, please do a version bump :)

PR #233, opened 2019-03-06
Modal csrf fix

allanchau:
This PR fixes up the modal script to pick up custom modals.

**Testing**

- [x] Check all custom modals with a `modal` class have the csrf input.

PR #235, opened 2019-03-08
Scripts

allanchau:
This PR updates the scripts tag to support data urls and some additional attribute that were previously missing.

**Related PRs*

- #233 

**Testing**

- [x] Start the test app and search the page source for `<script data-test="test"></script>`.

**Notes**

- I've updated the dev environment to start with `yarn start` instead of `yarn dev`.
- The reset/import path has been updated to `/reset`. The default route `/` doesn't run any scripts anymore.


docs/defaults.rst
```diff
@@ -156,6 +156,24 @@ You should use the existing array as the array that is resolved with the promise
 
 The script objects can contain an additional ``inHead`` boolean option to optionally load the script in the head tag.
 
+To create data attributes, you can add a ``dataAttributes`` property with a key being prefixed with ``data-``. For example::
```
smebberson: Please update to `...property with a key that will be prefixed with ``data-`` when output in HTML.`

lib/scripts.js
```diff
@@ -1,5 +1,29 @@
 'use strict';
 
+const parseDataAttributes = (scriptObj) => {
+
+    // Don't mutate the original object.
+    const script = Object.assign({}, scriptObj);
+
+    if (!script.dataAttributes) {
```
smebberson: I think this should be above 6. You're actually, needlessly, duplicating and returning a different object by doing it this way.

lib/scripts.js
```diff
@@ -1,5 +1,29 @@
...
+        return script;
+    }
+
+    const dataAttributes = Object.keys(script.dataAttributes);
```
smebberson: Can't this be written more cleanly using `map`?

test/lib-scripts.test.js
```diff
@@ -0,0 +1,69 @@
+'use strict';
+
+const {
+    parseDataAttributes,
+    parseScripts,
+} = require('../lib/scripts');
+
+test('it does not mutate the original script', () => {
```
smebberson: You could put another test here if you want to make sure it doesn't needlessly duplicate the object, kinda worthless, but would be interesting to validate my first comment.

views/partials/foot.jade
```diff
@@ -2,9 +2,9 @@ if scripts && scripts.length
 	each script in scripts
 		if !script.inHead
 			if script.content
-					script(async=script.async crossorigin=script.crossorigin defer=script.defer integrity=script.integrity src=script.src text=script.text type=script.type)!= script.content
+					script(async=script.async crossorigin=script.crossorigin defer=script.defer integrity=script.integrity nomodule=script.nomodule nonce=script.nonce referrerpolicy=script.referrerpolicy src=script.src text=script.text type=script.type)&attributes(script.dataAttributes || {})!= script.content
```
smebberson: I would prefer the `script.dataAttributes || {}` happen in JavaScript, not Jade. This will make it easier to move away from Jade.

views/partials/head.jade
```diff
@@ -14,6 +14,6 @@ head
         each script in scripts
             if script.inHead
                 if script.content
-                    script(async=script.async crossorigin=script.crossorigin defer=script.defer integrity=script.integrity src=script.src text=script.text type=script.type)!= script.content
+                    script(async=script.async crossorigin=script.crossorigin defer=script.defer integrity=script.integrity nomodule=script.nomodule nonce=script.nonce referrerpolicy=script.referrerpolicy src=script.src text=script.text type=script.type)&attributes(script.dataAttributes || {})!= script.content
```
smebberson: Same as above regarding `dataAttributes`...

lib/scripts.js
allanchau: `dataAttributes` and `parsedAttributes` are objects though, I'm not sure how we can use `map` without converting them between objects and arrays a few times.

I could just remove the `dataAttributes` variable and use `Object.keys(script.dataAttributes)` directly if that makes it easier to read?

lib/scripts.js
allanchau: @smebberson With the Jade change below I think we need to keep this here and add a `script.dataAttributes` default.

lib/util/parse.js
```diff
@@ -0,0 +1,44 @@
...
+
+    if (!newObj.dataAttributes) {
+
+        newObj.dataAttributes = {};
+
+        return newObj;
+
+    }
+
+    const dataAttributes = Object.keys(newObj.dataAttributes);
```
smebberson: @allanchau, how come you're not using `map` here directly on `newObj.dataAttributes`?

allanchau: @smebberson oh sorry, I might have accidentally closed the other convo. If we use map we would have to convert the array to an object later on anyway, seems like that would be a bit more complex than what I've done here?

smebberson: @allanchau, yeah, fair enough, makes sense.

lib/scripts.js
smebberson: Yep, makes sense.

PR #251, opened 2019-04-12
Temp record change fix

allanchau:
This PR is a temporary fix for issues with saving records when using the versions plugin.

**Setup**

- [ ] Map in this PR.

**Testing**

- [ ] Test saving records never fail with an alert.


smebberson:
@allanchau, can this be closed now?

PR #252, opened 2019-04-13
Security improvements

smebberson:
This PR includes a couple of security related improvements:

- Updates cookies to be HttpOnly
- Updates cookies to be SameSite (strict)
- Provides more flexibility on session and cookie settings (via defaults)

**Setup**

- [x] Pull down this PR

**Testing**

- [x] Execute `rm -rf ./node_modules/`.
- [x] Execute `yarn` to install latest modules.
- [x] Execute `yarn start` to start the development environment.
- [x] Visit http://localhost:8888/.
- [x] Remove all cookies for the domain.
- [x] Refresh the page.
- [x] Check the `connect.sid` cookie is HttpOnly.
- [x] Check the `connect.sid` cookie is SameSite (strict).
- [x] Check the `connect.sid` cookie content is signed.
- [x] Check the `X-XSS-Protection` header exists and is set to `1; mode=block`.
- [x] Check the `X-Powered-By` header doesn't exist.
- [x] Check you can log in and use the site as normal.
- [x] Log out.
- [x] Visit http://localhost:8888/admin/model/mtOrg/list.
- [x] Check the `linzReturnUrl` cookie is HttpOnly.
- [x] Check the `linzReturnUrl` cookie is SameSite (strict).
- [x] Check the `linzReturnUrl` cookie content is signed.
- [x] Log in.
- [x] Make sure you're redirected directly to http://localhost:8888/admin/model/mtOrg/list.


app/lib/session.js
```diff
@@ -3,10 +3,9 @@
...
+const linzDefaults = require('linz/lib/defaults');
 
-module.exports = session({
-    resave: false,
-    saveUninitialized: false,
-    secret: 'sessionsecret',
-    store: new RedisStore({ client: new Redis() }),
-});
+const store = new RedisStore({ client: new Redis() });
+const sessionOptions = { ...linzDefaults['session options'], ...{ store } };
```
allanchau: @smebberson I don't think we specify which node version we support anywhere, but won't this only work with node 8+?

smebberson: That code, yes. But it's just the example app. I've just added a `engine.node` entry to `package.json` accordingly.

smebberson: @allanchau, that okay?

package.json
```diff
@@ -38,7 +38,7 @@
     "chalk": "^2.4.1",
     "clone": "1.0.2",
     "connect-flash": "^0.1.1",
-    "cookie-parser": "1.0.x",
+    "cookie-parser": "1.4.4",
```
allanchau: Should we start to use the caret instead of pinning repo versions?
It can possibly reduce bloat for whoever implements Linz and the lockfile should install the exact same version we test anyway.

app/lib/session.js
allanchau: Yep sounds good.

package.json
smebberson: I think it's insecure to do so. There have been a few NPM security exploits lately that have been worsened and spread by `.x` and `caret` pinning so I'd prefer not to.

allanchau:
@smebberson In fact I would even update engine.node to support `>= 8`?

smebberson:
> @smebberson In fact I would even update engine.node to support `>= 8`?

@allanchau, on what?

allanchau:
> > @smebberson In fact I would even update engine.node to support `>= 8`?
> 
> @allanchau, on what?

In that last commit you did for the example app. I think the object spread was implemented in node 8.X?
I'm not sure if we are using anything that requires node 8.6 though.

smebberson:
@allanchau, it's actually 8.6

> > > @smebberson In fact I would even update engine.node to support `>= 8`?
> > 
> > 
> > @allanchau, on what?
> 
> In that last commit you did for the example app. I think the object spread was implemented in node 8.X?
> I'm not sure if we are using anything that requires node 8.6 though.

@allanchau, based on https://node.green/, it's 8.6 (it was under a flag in 8.2).

allanchau:
@smebberson All good now, I'll let you merge it in.

PR #254, opened 2019-05-09
Export fixes

allanchau:
This PR fixes:
- Exports to use the listQuery if provided.
- CSV issue when there is only 1 export field and it contains spaces.
- Adds the ability to split an object field into multiple columns in an export.

**Related PRs**

- #238
- #241 

**Testing**

- [x] Using the test app, reset the data `http://localhost:8888/reset`.
- [x] Export the `Object field` field on the user model.
- [x] Make sure you can see the `Object field` and 3 other `Object field X` fields which are the properties of the objectField.
- [x] Update L59 of `app/models/mtUser/formtools/form.js` 
to be false `'export': (val) => Promise.resolve([val, false]),` and create another export.
- [x] You should not see any of the property fields `Object field X`.
- [x] Update L59 of `app/models/mtUser/formtools/form.js` 
to be undefined `'export': (val) => Promise.resolve(val),` and create another export.
- [x] You should not see any of the property fields `Object field X`.

**Notes**

- Usually you would filter out the column header for the merge field, but I have left it in there to show the difference between the original and merged fields.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## Unreleased
+
+- Added the ability to split object fields over multiple columns in the default export.
+- Made the export api more generic so you can pass in any stream instead of a query.
+- Fixed the transform function to pass the correct parameters.
+
 ## v1.0.0-18.3.1 (18 April 2019)
 
 - Fixed concurrency plugin.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # CHANGELOG
 
-## Unreleased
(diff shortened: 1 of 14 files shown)


allanchau:
@smebberson Ready for you now.

allanchau:
@smebberson Yeah, that is what I meant by
> Usually you would filter out the column header for the merge field, but I have left it in there to show the difference between the original and merged fields.
I can filter it out if you want?

smebberson:
@allanchau, that's fine, as long as it can be programmed to not have those headers, that's fine. If you can just confirm that's a `yes`, then please do a version bump and release :)

PR #262, opened 2019-06-06
Export filter api

allanchau:
This PR exposes a new api to get filters from a post request.

**Testing**

- [x] Start the app and run an export, nothing should actually change.


smebberson:
@allanchau, could you please move these to the `linz.api.formtools.list` namespace?

lib/api/middleware/set-linz-model.js
```diff
@@ -2,14 +2,42 @@
...
+
+    // The param is config for config routes and model for others.
+    const { params = {} } = req;
+    const model = params.model || params.config;
+
+    if (model) {
+        return model;
+    }
+
+    const modelMatch = req.path.match(/(model\/\w*)/);
```
smebberson: @allanchau, please limit this to the models that Linz knows about only.

lib/router.js
```diff
@@ -46,7 +46,7 @@ exports.setupAdminRoutes = function () {
 
 exports.setupModelRoutes = function () {
 
-    router.use('/model/:model', linz.api.middleware.setLinzModel());
+    router.use('/model?/:model', linz.api.middleware.setLinzModel());
```
smebberson: @allanchau, why this change?

lib/api/middleware/set-linz-model.js
```diff
@@ -2,14 +2,42 @@
 
 const linz = require('../../../');
 
+/**
+ * Get the model name from a request.
+ * @param {Object} req The HTTP request object.
+ * @returns {String} The model name.
+ */
+const getModelFromRequest = (req) => {
```
smebberson: This isn't middleware. Please move it to `linz.api.model` and call it `fromRequest`.

lib/router.js
allanchau: @smebberson We need to be able to pick up custom routes too, eg `/model/testModel/action/export`. That won't be picked up by `'/model/:model'` in my testing.

smebberson: Do we though? I thought that was the whole point of the changes in `set-linz-model`, it would recognise the `/model/{model}/` pattern?

lib/api/middleware/set-linz-model.js
allanchau: @smebberson This expands upon `const model = req.params.model || req.params.config;` and is only intended to be used in this file. It is being exported so I could write some tests for the function.

lib/router.js
smebberson: This change means that `model` is optional? I don't see how that would work?

lib/api/middleware/set-linz-model.js
smebberson: Hmm. I can see what you're saying. You generally don't test stuff that isn't exposed though (good testing practices), so the fact that you're exposing it to test either means, we shouldn't have those tests, or we should expose it (and in this case move it to a proper location).

I think it's a good bit of function, why not have it expose in `linz.api.model`?

lib/router.js
allanchau: Yeah but the middleware still needs to be called? We could keep it as is and require the developer implementing it to call `linz.api.middleware.setLinzModel()` if you prefer?
That would mean all custom export routes need to call that assuming they want to filter based on the list filters though.

smebberson: If the custom export routes are mounted in `linz.app`, this middleware will always run without this `?` customisation....?

allanchau: I think that is the problem though, they aren't mounted in Linz.
Currently developers have to implement it themselves so `req.params.model` is undefined.

lib/api/middleware/set-linz-model.js
```diff
@@ -32,4 +60,7 @@ const setLinzModel = () => (req, res, next) => {
 
 };
 
-module.exports = setLinzModel;
+module.exports = {
+    getModelFromRequest,
```
smebberson: @allanchau, we shouldn't expose this if it's not intended for public use. Seeing as though `setLinzModel` is public, and it uses `getModelFromRequest`, can you just test via that?

allanchau:
@smebberson This should be ready for you again, the `getModelFromRequest` is no longer exposed and the test run more closely to how it would in a real application.

smebberson:
@allanchau, I'm happy with this now. Could you please version bump, launch and merge?

PR #266, opened 2019-07-16
Custom reset password path

allanchau:
This PR adds the ability to define custom password reset paths.

**Verification**

- [ ] Set the option `'admin forgot password path'`. Note the admin path is always prepended.

**Testing**

- [ ] Pull in this branch.
- [ ] Set the option `'admin forgot password path'` to something external. Note the admin path is not prepended.
- [ ] Remove the custom option. The admin path should be the same as before.


smebberson:
@allanchau, I think it would be better to just update the default to include `/admin` and then you can do away with the if statement? Also, if we're allowing updates `admin forgot password path`, do you think it would be good to do the same for `admin password reset path`?

allanchau:
@smebberson I've made those updates now.
I think we should leave `admin password reset path` for now. `admin forgot password path` was updated to allow specifying a completely different url which doesn't use any built in Linz routes. We need to revisit this at some point to allow changing the paths of all the admin routes, right now they are locked to `adminpath + path`.

PR #267, opened 2019-07-25
Export improvements

smebberson:
This PR updates the export functionality to prefer an inclusions list over an exclusions list.

## Setup

- [x] Pull down this PR.
- [x] Follow the steps to run Linz.

## Testing

- [x] Run the export for organisations (only the name field should be choose-able).
- [x] Run the export for users (password field should not be choose-able).
- [x] When the users export runs, the following deprecation warning should appear in the console:
```
DeprecationWarning: Export exclusions have been deprecated, use inclusions instead.
```


middleware/modelExport.js
```diff
@@ -162,151 +169,130 @@ module.exports = {
...
 
-                req.linz.model.listQuery(req, query, (listQueryErr, listQuery) => {
+                        if (listQueryErr) {
+                            return next(listQueryErr);
+                        }
 
-                    if (listQueryErr) {
-                        return next(listQueryErr);
-                    }
+                        const exclusions = exportObj.exclusions.split(',');
```
DinoLeung: i think this variable can be removed

smebberson: @DinoLeung, good pick up, I've just fixed that up.

PR #285, opened 2020-03-26
Customised passport

allanchau:
This PR adds the ability to customise the linz authenticate middleware.

**Testing**

- [x] Include this branch into your project.


app/app.js
```diff
@@ -19,23 +19,35 @@ const whoAmI = require('./routes/who-am-i');
...
 
         linz.init({
-            'express': app,
-            'options': {
+            express: app,
+            options: {
+                'login middleware': {
+                    get: [linz.api.middleware.login.get],
+                    post: [
+                        linz.api.middleware.authenticate('linz-local', {
```
smebberson, line 40: @allanchau, do you think this should be available at `linz.api.middleware.login.post`?

smebberson:
@allanchau, could you please also update/add to the documentation for this change?

app/app.js
allanchau: @smebberson I've moved this bit of code to `linz.api.middleware.login.post` now.

allanchau:
@smebberson I don't think so, it should work fine with all current projects. We just added the ability to customise the login.

PR #328, opened 2021-10-12
Sorted nav

allanchau:
This PR sorts the navigation items.

**Testing**

- [ ] Currently you should find navigation items are sorted by the db collection name. Using this pr they should now be sorted via the model name.

**Notes**

- You can still add navigation items wherever you like through the `navigationTransform` linz option.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "linz",
-    "version": "1.0.0-18.9.0",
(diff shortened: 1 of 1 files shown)


allanchau:
Hrmm prettier seems to have done its thing, the only changes are the sort(collator.compare) functions

PR #344, opened 2022-03-10
Fix for document array renderer

allanchau:
This PR fixes the document array renderer when viewing a model's versions.


smebberson (approved):
@allanchau, please launch.

PR #351, opened 2022-04-13
Allow aggregate

allanchau:
This PR adds the ability to pass an aggregate into the `getQuery` static function for a model.

**Notes**

- This PR also changes the way the `getCount` method works. It should now return an integer instead of returning a query and relying on linz to call query.count().


smebberson:
@allanchau, I'm closing this because we decided to go another path within the project itself.

